PR #22, opened 2023-03-14
Feature/#5 introduce configuration

### LINKED ISSUE #5: Introduce configuration
Introduce a configuration setup, most probably implemented by TypeSafe config.

Atum-Service project can be used for ideas, some version of TypeSafe config exists there and should be compatible.

## Design hints
Here, we are looking to provide some configuration defaults and external overrides per deployment needs.
This can perhaps be a form of `reference.conf` (internal defaults)+`application.conf` (externally provided) or also overriding individual fields external `-Dconfig.entry=value`.

Inspiration can be taken from https://github.com/AbsaOSS/atum-service -- it has a similar setup.

### List of users
~Optionally, configuration for defined set of users may be sketched out. (This especially makes sense when implementing after #7)~ moved to a separate issue #21

AnilPinarOzdemir:
BaseConfig & application.properties added.  

Closes #5


service/src/main/resources/application.properties
```diff
@@ -0,0 +1,5 @@
+server.port = 9090
+spring.application.name = login-gateway
+logingw.rest.config.some-key = BETA
+logingw.rest.config.alg-name = RS256
+logingw.rest.config.exp-time = 2
```
dk1844, line 5: Please add newline at the end

service/src/main/scala/za/co/absa/logingw/rest/config/BaseConfig.scala
```diff
@@ -0,0 +1,27 @@
...
+
+import org.springframework.boot.context.properties.{ConfigurationProperties, ConstructorBinding}
+
+@ConstructorBinding
+@ConfigurationProperties(prefix = "logingw.rest.config")
+class BaseConfig(
+                val someKey: String,
+                val algName: String,
+                val expTime: Int
+                )
```
dk1844, line 27:
```suggestion
class BaseConfig(
  val algName: String,
  val expTime: Int,
  val someKey: String = "example"
)
```

Please adjust the indentation.
Suggesting someKey to go last (can be more easily dropped later) and have a default value (does not have to be set where not needed).

dk1844 (commented):
Otherwise LGTM for now, we will extend in subsequent PRs.

AnilPinarOzdemir (commented):
New line added to application.properties
Indentation fixed in BaseConfig.scala

service/src/main/scala/za/co/absa/logingw/rest/config/BaseConfig.scala
```diff
@@ -0,0 +1,27 @@
...
+
+import org.springframework.boot.context.properties.{ConfigurationProperties, ConstructorBinding}
+
+@ConstructorBinding
+@ConfigurationProperties(prefix = "logingw.rest.config")
+class BaseConfig(
+  val algName: String,
+  val expTime: Int,
+  val someKey: String = "example"
+                )
```
dk1844, line 27:
```suggestion
class BaseConfig(
  val algName: String,
  val expTime: Int,
  val someKey: String = "example"
)
```

Almost :)

service/src/test/scala/za/co/absa/logingw/rest/service/JWTServiceTest.scala
```diff
@@ -19,13 +19,15 @@ package za.co.absa.logingw.rest.service
...
 import za.co.absa.logingw.model.User
+import za.co.absa.logingw.rest.config.BaseConfig
 
 import java.util
 import scala.util.Try
 
 class JWTServiceTest extends AnyFlatSpec {
 
-  private val jwtService: JWTService = new JWTService()
+  private val testConfig = new BaseConfig(algName = "RS256", someKey = null, expTime = 2)
```
dk1844, line 29:
```suggestion
  private val testConfig = new BaseConfig(algName = "RS256", expTime = 2)
```

`someKey` should not be necessary now, when it has a default value, I assume.

service/src/main/scala/za/co/absa/logingw/rest/service/InfoService.scala
```diff
@@ -18,10 +18,11 @@ package za.co.absa.logingw.rest.service
...
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.stereotype.Service
+import za.co.absa.logingw.rest.config.BaseConfig
 
 @Service
-class InfoService @Autowired()() {
+class InfoService @Autowired()(baseConfig: BaseConfig) {
   def getInfoMessage: String = {
-    s"Basic info message to be here." // todo service version, some parts of configuration perhaps - Issue #15
+    s"Basic info message to be here. '${baseConfig.someKey}'" // todo service version, some parts of configuration perhaps - Issue #15
```
TheLydonKing, line 26: The info service should indicate what ${baseConfig.someKey} means to the user.

The Service is currently just a shell, so not a big deal at the moment.

service/src/test/scala/za/co/absa/logingw/rest/service/InfoServiceTest.scala
```diff
@@ -32,6 +33,6 @@ class InfoServiceTest extends AnyFlatSpec with Matchers {
   new TestContextManager(this.getClass).prepareTestInstance(this)
 
   "InfoService" should "give expected test message" in {
-    infoService.getInfoMessage shouldEqual "Basic info message to be here."
+    infoService.getInfoMessage shouldEqual "Basic info message to be here. 'BETA'"
```
TheLydonKing, line 36: Tests Should be able to parse the value from the config as well instead of hardcoding the value. This ensures that we don't have to change the tests every time we change the config and also helps if we introduce different config profiles for different environments.

dk1844 (approved):
Sure, good enough as a starting point for future extensions.

@AnilPinarOzdemir, please respond appropriately to the comments from above, too.

service/src/main/scala/za/co/absa/logingw/rest/config/BaseConfig.scala
```diff
@@ -0,0 +1,27 @@
...
+ * limitations under the License.
+ */
+
+package za.co.absa.logingw.rest.config
+
+import org.springframework.boot.context.properties.{ConfigurationProperties, ConstructorBinding}
+
+@ConstructorBinding
+@ConfigurationProperties(prefix = "logingw.rest.config")
+class BaseConfig(
```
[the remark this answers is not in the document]
AnilPinarOzdemir: Changed BaseConfig into a case class. The implementation of potential new specific configs would be creating new case classes for each SpecificConfig.

service/src/main/scala/za/co/absa/logingw/rest/service/InfoService.scala
AnilPinarOzdemir: As you mentioned , since the service is currently just a shell, this is just a showcase. Can be updated on next feature PRs.

service/src/test/scala/za/co/absa/logingw/rest/service/InfoServiceTest.scala
AnilPinarOzdemir: Test updated as you mentioned above. With the latest PR, now both tests are able to parse the value from the config instance.

PR #37, opened 2023-05-09
Feature/36 externalizing config

### LINKED ISSUE #36: Config externalization
## Background
Currently, we have a builtin `application.yaml` config with example data - while it is doable to override it, it would be cleaner to actually provide just an example of it and the user will be forced to create his own - they made decisions that the example content is acceptable for some simple usages, but they will have to take action and possibly will be forced to put a thought to the fact whether the content they have been exposed to fits their need.

## Feature
Rename the current config file to indicate it is not a default for normal usage
Provide or document ways to provide user config externally.

dk1844:
This PR bring some updates forcing the user to provide their own `application.yaml` config.
All is described in the README update.

For the simplest test-run, you can either just copy&rename `example.application.yaml` -> `application.yaml` and it will be picked up by default. Other options in the readme and at [Spring boot - Externalized Configuration](https://docs.spring.io/spring-boot/docs/current/reference/html/features.html#features.external-config).

Closes #36.


dk1844:
> Info and Health endpoint tests seem to be failing. Seems to be failing due to config according to the logs.

Thanks, thanks for noticing before I did. Yes, that test relies on an application config being present - so I have added one for tests only.

jakipatryk (approved):
LGTM (just checked code).

PR #38, opened 2023-05-12
Feature/#5 introduce configuration

AnilPinarOzdemir:
#5 
Tests that make sure all parameters in application.yaml are initialized  by config classes properly are added in ConfigsTest
BaseConfigTest refactored to ConfigsTest


dk1844 (approved):
LGTM

PR #39, opened 2023-05-18
feature/#15 info endpoint

TheLydonKing:
- Removed old Info Endpoint
- Removed old Info Endpoint Unit Test
- Added Actuator Info Endpoint
- Add Example git.properties file
- Added Info Endpoint Unit Test
- Updated Readme for Actuator Info Endpoint
- Added tooling for user to generate their own git.properties file automatically


README.md
```diff
@@ -102,3 +102,26 @@ Local JMX is currently enabled on the project. In order to utilize it please fol
...
+Springboot Actuator is enabled for this project. This enables an Info endpoint that can be populated with various information that may be
+useful to the troubleshooting and usage of the application. The endpoint can be accessed via the following url:  `http://localhost:port/actuator/info`.
+The information types available are as follows:
+
+| ID     | Usage                                                        |
+|--------|--------------------------------------------------------------|
+ | build  | Exposes build information                                    |
+| env    | Exposes any property marked under "info" in the config       |
+ | git    | Exposes git related information from the git.properties file |
+| java   | Exposes Java runtime information                             |
```
dk1844, line 116: I think the table does not make sense on its own here. For example, the user cannot deduce that for java-related info to be included, they should add `management.info.java.enabled=true` (which works).

Either remove the table altogether or extend it by explaining what to do in the config to gain the extra info.

I'd probably remove it and maybe extend the example config yaml with the `java` info as a comment the user can add.

README.md
```diff
@@ -102,3 +102,26 @@ Local JMX is currently enabled on the project. In order to utilize it please fol
...
+
+In the given example config ("example.application.yaml"), the env and git properties are enabled. 
+An example git.properties file has also been included ("example.git.properties"), simply rename it to "git.properties" in order to make use of it.
```
dk1844, line 119: This works. But how do you envision the process actually expected to work - where is the `git.properties` file supposed to come from? Should it be included internally - when the project is ran from source or externally when released? or both?

service/src/main/resources/example.git.properties
```diff
@@ -0,0 +1,4 @@
+git.commit.id=git_id
+git.commit.time=5/15/2023
+git.commit.user.email=exampleuser@org.com
+git.commit.message.full=Added Git Properties
```
dk1844, line 4: missing new line

service/src/test/resources/application.yaml
```diff
@@ -28,21 +28,32 @@ spring:
...
-        include: "health"
+        include:
+          - "health"
+          - "info"
   endpoint:
     health:
       show-details: "never"
+
+info:
+  test: "This is a test value"
```
dk1844, line 59: missing new line

dk1844 (commented):
Very nice, I also tried it, works well. Just posted a couple of comments.

README.md
TheLydonKing: Done so. Removed table and added comments to example config on what each property does

README.md
TheLydonKing: Added tooling for user to generate git.properties file. Maven and Gradle have plugins that handle this automatically but couldn't find a solution for SBT, so just made one myself.

README.md
```diff
@@ -106,22 +106,18 @@ Remote JMX is also an option and may be enabled with some config changes in the
...
 
 If you wish to change what is shown here, you can do so by changing the fields and attributes in the application file. 
-More info on the Actuator Info Service can be found here: https://reflectoring.io/spring-boot-info-endpoint/
\ No newline at end of file
+More info on the Actuator Info Service can be found here: https://reflectoring.io/spring-boot-info-endpoint/
+
+## git.properties
+An example git.properties file has also been included ("example.git.properties"), simply rename it to "git.properties" in order to make use of it in the info endpoint.
+If you wish to generate an accurate git.properties file, you can do so by setting the "logingw.rest.config.git-properties" to true.
+This will generate a new file with updated git information on application startup. You may need to set the property to false and restart the application once the file has been generated in order for it to be visible on the info endpoint.
```
dk1844, line 123: It is a bit unfortunate that the git info needs two starts (first to generate, second to load). I suggest making an issue about that and linking it here if this cannot be fixed easily/quickly.

service/src/main/resources/example.application.yaml
```diff
@@ -6,6 +6,7 @@ logingw:
       alg-name: "RS256"
     config:
       some-key: "BETA"
+      git-properties: false
```
dk1844, line 9: This key would deserve a comment on how it works (this is to generate on startup, but will not be loaded unless restarted)

service/src/main/scala/za/co/absa/logingw/rest/config/GitConfig.scala
```diff
@@ -0,0 +1,60 @@
...
+case class GitConfig(gitProperties: Boolean) {
+
+  @PostConstruct
+  def init(): Unit = {
+    if(gitProperties)
+    this.GitPropertiesGenerator.generateGitProperties()
+  }
+
+  object GitPropertiesGenerator {
+    def generateGitProperties(): Unit = {
```
dk1844, line 36: pretty cool. Works on Windows if you have `git` available.

dk1844 (approved):
LGTM (read, tested).

service/src/main/scala/za/co/absa/logingw/rest/config/GitConfig.scala
TheLydonKing: Added a note in the Read Me and Config about this. Thanks for reminding me

service/src/main/resources/example.application.yaml
TheLydonKing: Did So now

README.md
TheLydonKing: Will do, will have to think it through. Thinking it might be better to just add it as either a test or it's own service in the project.

Issue #40, opened 2023-05-25
Improve upon the current git.properties generation


TheLydonKing:
## Background
A feature has been added in #39 to the project for users to generate git information to display on the info endpoint. More elegant solutions exist for Maven and Gradle but for SBT there don't seem to be any predefined solutions (Or at least nothing that I could find). 

## Feature
The solution at the moment isn't too elegant as it requires the app to be restarted after the initial file generation. This issue is to improve upon the current solution.

## Proposed Solution
Solution Ideas:
1.  **Try and generate the file at compile time rather than runtime?** I'm still a beginner with SBT but it seems like you can generate some sources and files at compile time with SBT? (https://www.scala-sbt.org/1.x/docs/Howto-Generating-Files.html) Please correct me if I'm wrong.
2. Optional extra: Separate the tool into its own executable within the project.
3. Optional extra: Have the tool run as a test by users.
4. Please propose a solution if you have any ideas.

PR #41, opened 2023-05-31
#40 Improve on git.properties generation

TheLydonKing:
- Added Custom Info Endpoint that is used when file generation is turned on. Removes old requirement of restarting the service for information to reflect.
- Added Test for Custom Endpoint
- Added file generation to a test so that user may generate the file before packaging the application.
- added Try Catches to the code as there are dependencies that could cause failure.
- Changes reflected in ReadMe


dk1844 (changes requested):
I am afraid that his PR does not solve the primary underlying problem - the app had to be restarted to automatically get the `git.properties` generated (provided the config has the `git-properties: true` flag set).

This PR removes that option altogether which I find actually worse than it was (it was a tiny bit harder to do, but at least it did not require the user unfamiliar with the code to make changes).

I am not against having this option in tests, too, but otherwise, I do feel that the original behavior (alas with two starts until figured out to be just one) should be kept. Please bring that part back.

dk1844 (commented):
Also, what about renaming the config to better communicate the intention?
Perhaps `git-properties: true` -> `generate-git-properties: true`?

service/src/main/scala/za/co/absa/logingw/rest/config/GitConfig.scala
```diff
@@ -18,42 +18,60 @@ package za.co.absa.logingw.rest.config
...
+    if(generateGitProperties)
+      GitPropertiesGenerator.generateGitProperties()
   }
+}
+
+case class GitProperties(branch: String, commitId: String, commitTime: String)
+
+object GitPropertiesHolder {
+  var gitProperties: GitProperties = _
+}
```
dk1844, line 40: This looks a unusal -- it does not seem very Scala idiomatic. Is there any other way to do it without the `var`? Perhaps some sort of dependency injection?

build.sbt
```diff
@@ -31,6 +31,7 @@ lazy val commonJacocoReportSettings: JacocoReportSettings = JacocoReportSettings
 lazy val commonJacocoExcludes: Seq[String] = Seq(
 //    "za.co.absa.logingw.model.User*", // class and related objects
 //    "za.co.absa.logingw.rest.config.BaseConfig" // class only
+      "src/main/scala/za/co/absa/logingw/rest/config/GitConfig.scala"
```
dk1844, line 34: why do you exclude this class from JaCoCo?

README.md
```diff
@@ -119,8 +119,20 @@ More info on the Actuator Info Service can be found here: https://reflectoring.i
...
+   In order to run the test the line (line 30) that reads as:
+   ```
+   ignore should "generate git.properties file" in {
+   ```
+   should be changed to:
+   ```
+   "This function" should "generate git.properties file" in {
+   ```
+   once this is done, running or debugging the test will generate a git.properties file to be used for the info endpoint.
+
```
dk1844, line 136: Suggesting:

If you wish to generate an accurate git.properties file, you can do so in 2 ways:

1) Setting the `logingw.rest.config.generate-git-properties` to `true`.
   This will generate a new file with updated git information on the application startup.
2) By manually adjusting and running the test in `za.co.absa.logingw.rest.actuator.toolingGitPropertiesGenerator`. 

   In order to run the test the line (line 30) that reads as:
   ```scala
   ignore should "generate git.properties file" in {
   ```
   should be changed to:
   ```scala
   "This function" should "generate git.properties file" in {
   ```
   Once this is done, running or debugging the test will generate a `git.properties` file to be used for the info endpoint.

dk1844 (commented):
I have tested in on my machine, works well.

service/src/main/resources/example.application.yaml
```diff
@@ -9,7 +9,9 @@ logingw:
       #Generates git.properties file for use on info endpoint.
       #Application will have to be restarted after file generation for information to be visible.
       #Recommended to set the property to false after file has been generated.
```
dk1844, line 11: This is outdated now. There are two properties now to describe

service/src/main/scala/za/co/absa/logingw/rest/config/GitConfig.scala
```diff
@@ -18,42 +18,72 @@ package za.co.absa.logingw.rest.config
...
-      val gitCommitId = sys.process.Process("git rev-parse HEAD").lineStream_!.headOption.getOrElse("unknown")
-      val gitCommitTime = dateFormat.format(sys.process.Process("git show -s --format=%ct HEAD").lineStream_!.headOption.map(_.toLong * 1000).getOrElse(0L))
-      val gitAuthor = sys.process.Process("git show -s --format='%ae'").lineStream_!.headOption.getOrElse("unknown")
-      val content =
-        s"""|git.branch=$gitBranch
-            |git.commit.message.full=$gitCommitMessage
-            |git.commit.id=$gitCommitId
-            |git.commit.time=$gitCommitTime
-            |git.commit.user.email=$gitAuthor
-            |""".stripMargin
```
dk1844, line 49: I see that some of the fields were omitted, but that is probably ok -- the newer version better conforms to the git info in the _simple_ form provided by the maven plugin 👍

dk1844 (approved):
LGTM aside from the outdated config desc (see below)

dk1844 (approved):
LGTM

PR #51, opened 2023-06-21
Feature/49 service dockerization

### LINKED ISSUE #49: Service Dockerization
## Background
In order to run in docker (EKS/Fargate, elsewhere), we should provide a docker image file to be usable for such a purpose.

## Feature
Create a dockerfile runnable in Docker and attempt to run it in Absa's AWS Cloud (presumably EKS or Fargate)
Make the Dockerfile configurable in order for custom ARGs to be used to tailor the image to one's needs (e.g. Absa's cloud deployments with specific settings)

dk1844:
Basic docker image.

Config is to be externally defined via docker's 
```
docker build -t absaoss/login-service:latest `
--build-arg BUILD_PROXY=http://proxy.ctonetwrkprod.aws.dsarena.com:3128 `
--build-arg BASE_IMAGE=artifacts.bcp.absa.co.za/docker.io/tomcat:9.0.62-jdk8-corretto `
--build-arg CONFIG=./src/main/resources/local.application.yaml `
--build-arg LDAP_CERT=./src/main/resources/absa-ldap.pem `
--build-arg SSL=true `
--build-arg SSL_DNAME="CN=*.ctodataengprod.aws.dsarena.com, OU=ursa, O=absa, L=Johannesburg, ST=Gauteng, C=za" .
```

The intention is for the Dockerfile to be called in a company-private pipeline where the image creation is parametrized by the above and used further.

Closes #49


lsulak (commented):
If this is supposed to be cloud-use only, then I would assume that this will be deployed and used in AWS ECR (Elastic Code Registry).

In that case, I guess that the next step is to:
1. Add Terraform code for the ECR resource and additional configuration needed (like AWS resource labels etc)
2. Add the Docker build & deployment to the CICD.

But this is OSS, so it's gonna be in a different repo?

service/Dockerfile
```diff
@@ -0,0 +1,45 @@
...
+ENV http_proxy=$BUILD_PROXY
+ENV https_proxy=$BUILD_PROXY
+ENV HTTP_PROXY=$BUILD_PROXY
+ENV HTTPS_PROXY=$BUILD_PROXY
+
+# uncomment and add packages you would need to install via yum if any
+#RUN yum -y update
+#RUN yum -y install htop procps
+
+EXPOSE 8080
```
lsulak, line 42: Tomcat is known for using the default port 8443 for SSL HTTPS traffic. Shouldn't it be also added here, not just 8080? It's like that also in Unify project, see its Dockerfile.

That brings me to this (also in the Unify project):

```
# A self-seigned certificate for HTTPS
RUN keytool -genkeypair -keyalg RSA -alias tomcat -keysize 2048 \
    -dname "CN=*.ctodataengprod.aws.dsarena.com, OU=ursa, O=absa, L=Johannesburg, ST=Gauteng, C=za" \
    -validity 365 -storepass changeit -keystore /usr/local/tomcat/conf/selfsigned.p12 -storetype PKCS12
```

So this is for HTTPS - or do you want to deal with this detail later / it's not needed here (yet?)?

service/Dockerfile
```diff
@@ -0,0 +1,45 @@
+# LS docker file
+
+# Tomcat OS base image
+FROM artifacts.bcp.absa.co.za/docker.io/tomcat:9.0.62-jdk8-corretto
+
+LABEL org.opencontainers.image.authors="ABSA OSS"
+LABEL org.opencontainers.image.title="Login service"
+LABEL org.opencontainers.image.base.name="absaoss/login-service:latest"
```
lsulak, line 8: At least those 2 can be perhaps parametrized? If this is supposed to be more like company-wide image, not strictly related to LS.

service/Dockerfile
```diff
@@ -0,0 +1,45 @@
+# LS docker file
+
+# Tomcat OS base image
+FROM artifacts.bcp.absa.co.za/docker.io/tomcat:9.0.62-jdk8-corretto
```
lsulak, line 4: I guess that this is our hardened image, and there is absolutely nothing sensitive in it? (Since this is OSS repo)

service/Dockerfile
dk1844: Good idea, thanks, will do.

service/Dockerfile
dk1844: Good question, I have to find out.

dk1844: Replaced with configurable image, the default being vanilla tomcat image.

service/Dockerfile
dk1844: I have done exactly that, but optionally. so with docker build, one can either use
 - `--build-arg SSL=false` (same as until now - just HTTP @ 8080), or
 - `--build-arg SSL=true --build-arg SSL_DNAME="CN=*.abc.mydomain.com, OU=myproject, O=mycompany, L=Johannesburg, ST=Gauteng, C=za"` to get also HTTPS @ 8443.

service/Dockerfile
dk1844: Replaced with just `authors=ABSA` in the end.

service/Dockerfile
```diff
@@ -0,0 +1,78 @@
...
+#RUN yum -y install htop procps
+
+# --- SSL=true image specifics ---
+FROM base AS base-ssl-true
+ENV SSL_ENABLED=true
+RUN echo "This stage sets SSL=$SSL_ENABLED"
+# Copy Tomcat properties (to enable SSL)
+COPY ./src/main/resources/tomcat-ssl/server.xml /usr/local/tomcat/conf/
+
+ARG SSL_DNAME="CN=*.my.example.com, OU=project1, O=yourcompany, L=Johannesburg, ST=Gauteng, C=za"
```
lsulak, line 58: I would just move all ARGs to the beginning of the Dockefile unless there is a good reason not to..?

lsulak: (but I will leave it up to you because I understand that you kind of want to put it close to the SSL related code)

service/Dockerfile
```diff
@@ -0,0 +1,78 @@
...
+# Tomcat OS base image, feel free to use another (e.g. your hardened one)
+ARG BASE_IMAGE=tomcat:9-jdk8-corretto
+FROM $BASE_IMAGE as base
+
+LABEL org.opencontainers.image.authors="ABSA"
+
+ARG BUILD_PROXY
+ARG CONFIG=./src/main/resources/example.application.yaml
+
+ENV LS_WAR_PATH=./target/scala-2.12
```
lsulak, line 31: maybe even 2.12 can be somehow parametrized, but probably good enough for now

service/Dockerfile
```diff
@@ -0,0 +1,78 @@
...
+FROM base AS base-ssl-false
+ENV SSL_ENABLED=false
+RUN echo "This stage sets SSL=$SSL_ENABLED"
+
+EXPOSE 8080
+CMD ["catalina.sh", "run"]
+
+# --- Final image assembly ---
+FROM base-ssl-${SSL} AS final
+RUN echo "Running final stage with SSL_ENABLED=$SSL_ENABLED."
```
lsulak, line 78: since this is a multi-stage docker build, don't you need to copy some app/config files into the last build(s) ? I.e. after the last FROM (actually two of them obviously can be 'final')

E.g. these files 
 
```
# Copy Spring application properties
COPY $CONFIG /opt/application.yaml

COPY $LS_WAR_PATH/*.war /usr/local/tomcat/webapps/$LS_WAR
```

I haven't tried it though, just did the review. If you want me to build it on my Mac, please tell me

lsulak: (If I understood and remember multi-stage docker builds correctly, every FROM is a new layer and doesn't contain artifacts from the previous layers - ENV vars and files?)

service/Dockerfile
dk1844: The thing here is that this arg is not used if SSL=false

service/Dockerfile
dk1844: I think your comment would be correct in the case of multi-stage docker build that do not use each other as bases (as explained at e.g. https://www.baeldung.com/linux/docker-copy-variables-stages) - there, the extra copying and redefinitions of ARGs would be needed (while ENV being used throughout the whole file.

However, here, the `base` image is used via `FROM` in subsequent images and so on, creating a chain, so no extra copying or ARG redefinition should be needed. If it was needed and missing, the solution would not work properly and would the `final` image would be oblivious to ARGs specified for the `base` image -- but it works correctly, thus serving as evidence of no need of copy/ARG-redef in this case. It is somehow outlined at https://stackoverflow.com/questions/57577265/which-dockerfile-instructions-are-inherited-in-deriving-images where such an _image inheritance_ is discussed.

service/Dockerfile
dk1844: I am not planning on doing this presently, if in the future we support different Scala version, we can go for this option.

service/Dockerfile
lsulak: Okay, I didn't now about these particular inheritance capabilities. Cool, thanks :)

lsulak (approved):

lsulak:
LGTM, but I read a code and didn't build it

dk1844:
Known issues at the moment: 

- LDAP integration fails in docker (most probable cause = cacerts does not include LDAPS's public cert - not trusted)
- actuator endpoint relies on git being present on the machine

service/src/main/resources/tomcat-ssl/server.xml
```diff
@@ -0,0 +1,185 @@
...
+  <Listener className="org.apache.catalina.core.ThreadLocalLeakPreventionListener" />
+
+  <!-- Global JNDI resources
+       Documentation at /docs/jndi-resources-howto.html
+  -->
+  <GlobalNamingResources>
+    <!-- Editable user database that can also be used by
+         UserDatabaseRealm to authenticate users
+    -->
+    <Resource name="UserDatabase" auth="Container"
```
jakipatryk, line 41: What is this thing for?

service/Dockerfile
```diff
@@ -0,0 +1,88 @@
...
+
+# --- Base image part (common for SSL true|false) ---
+FROM $BASE_IMAGE as base
+
+# Provide your proxy if needed, e.g. http://my.proxy.examle.com:3128
+ARG BUILD_PROXY
+# Override of the example application config is possible
+ARG CONFIG=./src/main/resources/example.application.yaml
+# Provide path to LDAP cert in PEM format
+ARG LDAP_CERT
```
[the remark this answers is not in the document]
dk1844: Good idea, `ARG LDAP_SSL_CERT` used.

service/src/main/resources/tomcat-ssl/server.xml
dk1844: I don't know. The whole file is a default from fresh tomcat, the only change there is the the connector:
```xml
    <!-- This connector is the only change in otherwise original server.xml from image -->
    <Connector port="8443" protocol="HTTP/1.1" SSLEnabled="true"
               maxThreads="150" scheme="https" secure="true"
               keystoreFile="/usr/local/tomcat/conf/selfsigned.p12" keystorePass="changeit"
               clientAuth="false" acceptCount="100"/>
```

jakipatryk (approved):

PR #52, opened 2023-06-26
Feature/#28 list of customisable auth providers

TheLydonKing:
- Added the ability to choose which authentication methods are enabled and what order they are used in
- Added Validation for these methods to ensure that they are complete if enabled or to ensure that at least 1 is enabled
- Added Tests
- Added instruction in readme

Please let me know if you are fine with this implementation or if there is something a bit more user friendly.


README.md
```diff
@@ -50,6 +50,27 @@ It is available for download while running the service at `http://localhost:port
...
+The Login Service allows users to select which authentication providers they would like to use
+as well as the order in which the authentication methods get prioritized.
+
+This is done by setting a number (Greater than 0) on the "enable" tag for each auth method.
+the number used indicates the order in which you would like to use the method.
+
+For Example:
+The 2 methods currently enabled are config specified users:
+```
+logingw.rest.users.enable = 1
```
jakipatryk, line 63: Can `users` config be prefixed with `auth.` as in LDAP case?

README.md
```diff
@@ -50,6 +50,27 @@ It is available for download while running the service at `http://localhost:port
...
+```
+and ldap:
+```
+logingw.rest.auth.ad.ldap = 2
```
jakipatryk, line 67: `.enable` suffix is missing

jakipatryk (commented):
Just checked the code quickly for now.

README.md
TheLydonKing: Ah, yeah, makes sense to me to put it all under one tag. Will do

README.md
TheLydonKing: Woops, will fix

service/src/main/resources/example.application.yaml
```diff
@@ -17,6 +17,10 @@ loginsvc:
 
     # Users (config-defined)
     users:
+      #Auth Protocol
```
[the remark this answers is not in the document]
TheLydonKing: Will do

service/src/main/resources/example.application.yaml
```diff
@@ -34,6 +38,10 @@ loginsvc:
     auth:
       ad:
         ldap:
+          #Auth Protocol
```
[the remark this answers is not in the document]
TheLydonKing: Will do

service/src/main/scala/za/co/absa/loginsvc/rest/AuthManagerConfig.scala
```diff
@@ -18,36 +18,49 @@ package za.co.absa.loginsvc.rest
...
+    }
+
+    if (adLDAPConfig != null) {
+      val adLDAPAuthProvider = new ActiveDirectoryLDAPAuthenticationProvider(adLDAPConfig)
+      providerMap += (adLDAPConfig.enable -> adLDAPAuthProvider)
+    }
+
+    val orderedProviders = providerMap
+      .result()
+      .filter(x => x._1 > 0)
```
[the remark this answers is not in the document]
TheLydonKing: Agreed, since just excluding the config does the same thing as enabled = 0, makes it a bit simpler to use, will implement

service/src/main/scala/za/co/absa/loginsvc/rest/config/auth/UsersConfig.scala
```diff
@@ -25,31 +25,36 @@ import javax.annotation.PostConstruct
 @ConstructorBinding
 @ConfigurationProperties(prefix = "loginsvc.rest.users")
 case class UsersConfig(
-  knownUsers: Array[UserConfig],
-) extends ConfigValidatable {
+                        knownUsers: Array[UserConfig], enable: Int
```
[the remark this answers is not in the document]
TheLydonKing: ... not sure. It does look odd but that's how my intelliJ is formatting it.

Let me make it look a bit neater.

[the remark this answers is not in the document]
jakipatryk: Ok sorry I just checked that this project actually doesn't have ScalaFMT config. So can be as it is, maybe we will add ScalaFMT in the future and do reformatting automatically on commit/push. Not important for now.

service/src/main/scala/za/co/absa/loginsvc/rest/AuthManagerConfig.scala
```diff
@@ -18,36 +18,49 @@ package za.co.absa.loginsvc.rest
...
-  // TODO make it autowired but only if in config AD LDAP provider is set up (#28)
-  usersConfig: UsersConfig,
-  adLDAPConfig: ActiveDirectoryLDAPConfig
-){
+class AuthManagerConfig{
+
+  @Autowired(required = false)
+  private var usersConfig: UsersConfig = _
+  @Autowired (required = false)
+  private var adLDAPConfig: ActiveDirectoryLDAPConfig = _
```
dk1844, line 34: why vars?

I'd guess
```scala
  @Autowired(required = false)
  val usersConfig: UsersConfig = null
  @Autowired(required = false)
  val adLDAPConfig: ActiveDirectoryLDAPConfig = null
```
could be used.

service/src/main/scala/za/co/absa/loginsvc/rest/AuthManagerConfig.scala
```diff
@@ -18,36 +18,49 @@ package za.co.absa.loginsvc.rest
...
+      providerMap += (adLDAPConfig.order -> adLDAPAuthProvider)
+    }
+
+    val orderedProviders = providerMap
+      .result()
+      .filter(x => x._1 > 0)
+      .toList
+      .sortBy(_._1)
+      .map { case (_, provider) => provider }
+      .toArray
```
dk1844, line 58: SortedMap can be used and it has default ordering by keys and method `.values`. Hope it works for Scala 2.12 as well: https://www.scala-lang.org/api/2.13.6/scala/collection/immutable/SortedMap.html

So maybe a little less ordering juggling.

dk1844 (commented):
I have tried it, it seems to work correctly.

A possible point of improvement: It would be better if on startup, the app would log which auth providers are actually enabled (ideally with the order given).

service/src/main/scala/za/co/absa/loginsvc/rest/AuthManagerConfig.scala
```diff
@@ -18,36 +18,49 @@ package za.co.absa.loginsvc.rest
...
+
+    if (usersConfig != null && usersConfig.order > 0) {
+      val configUsersAuthProvider = new ConfigUsersAuthenticationProvider(usersConfig)
+      providerMap += (usersConfig.order -> configUsersAuthProvider)
+    }
+
+    if (adLDAPConfig != null && adLDAPConfig.order > 0) {
+      val adLDAPAuthProvider = new ActiveDirectoryLDAPAuthenticationProvider(adLDAPConfig)
+      providerMap += (adLDAPConfig.order -> adLDAPAuthProvider)
+    }
```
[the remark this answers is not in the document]
TheLydonKing: abstracted this to single function.

service/src/main/scala/za/co/absa/loginsvc/rest/AuthManagerConfig.scala
TheLydonKing: Implemented

service/src/main/scala/za/co/absa/loginsvc/rest/AuthManagerConfig.scala
TheLydonKing: Using Sortedmaps now, Code should be A LOT simpler now

service/src/main/scala/za/co/absa/loginsvc/rest/AuthManagerConfig.scala
```diff
@@ -21,46 +21,41 @@ import org.springframework.context.annotation.{Bean, Configuration}
...
   }
+
+  private def createProviders(configs: Array[DynamicAuthOrder]): SortedMap[Int, AuthenticationProvider] = {
+    val resultsMap = SortedMap.empty[Int, AuthenticationProvider] ++ configs.filter(_.order > 0)
+      .map {
+        case config@(c: UsersConfig) => (config.order, new ConfigUsersAuthenticationProvider(c))
+        case config@(c: ActiveDirectoryLDAPConfig) => (config.order, new ActiveDirectoryLDAPAuthenticationProvider(c))
+        case _ => throw new IllegalStateException("unsupported config type")
+      }
+    resultsMap.range(1, 999)
```
dk1844, line 66: what is the point?
if someone gives order 1234, it would get removed without the user knowing.
configs with `order <= 0` are ignored anyway, so perhaps this is not needed at all?

service/src/main/scala/za/co/absa/loginsvc/rest/AuthManagerConfig.scala
```diff
@@ -18,36 +18,44 @@ package za.co.absa.loginsvc.rest
...
+    orderedProviders.foreach(auth => authenticationManagerBuilder.authenticationProvider(auth._2))
+    authenticationManagerBuilder.build
   }
 
+  private def createProviders(configs: Array[DynamicAuthOrder]): SortedMap[Int, AuthenticationProvider] = {
+    val resultsMap = SortedMap.empty[Int, AuthenticationProvider] ++ configs.filter(_.order > 0)
+      .map {
+        case config@(c: UsersConfig) => (config.order, new ConfigUsersAuthenticationProvider(c))
+        case config@(c: ActiveDirectoryLDAPConfig) => (config.order, new ActiveDirectoryLDAPAuthenticationProvider(c))
+        case _ => throw new IllegalStateException("unsupported config type")
```
dk1844, line 64:
```suggestion
        case c: UsersConfig => (c.order, new ConfigUsersAuthenticationProvider(c))
        case c: ActiveDirectoryLDAPConfig => (c.order, new ActiveDirectoryLDAPAuthenticationProvider(c))
        case other => throw new IllegalStateException(s"unsupported config type $other")
```

I think there is no need for `config@(c: ...)`, can be simplified as ⬆️

service/src/main/scala/za/co/absa/loginsvc/rest/AuthManagerConfig.scala
```diff
@@ -18,36 +18,44 @@ package za.co.absa.loginsvc.rest
...
-        new ActiveDirectoryLDAPAuthenticationProvider(adLDAPConfig)
-      )
-      .build
+
+    val authenticationManagerBuilder = http.getSharedObject(classOf[AuthenticationManagerBuilder]).parentAuthenticationManager(null)
+    val configs: Array[DynamicAuthOrder] = Array(usersConfig, adLDAPConfig)
+    val orderedProviders = createProviders(configs)
+
+    if(orderedProviders.isEmpty)
+      throw new Exception("No authentication method enabled in config")
```
dk1844, line 49: Just thinking -- here (and maybe below with the IllegalStateException), what about using an appropriate Exception type? Perhaps reusing the ConfigValidationException? Does it fit or do we need to create another type?

README.md
```diff
@@ -50,6 +50,28 @@ It is available for download while running the service at `http://localhost:port
...
+```
+loginsvc.rest.auth.provider.users.order = 1
+```
+and ldap:
+```
+loginsvc.rest.auth.provider.ldap.order = 2
+```
+In the above example, both methods are enabled with the config specified users taking priority over Ldap.
+
+In order to disable an authentication protocol, set the `order` property to `0`
```
dk1844, line 71: I am unsure if this is a good idea or not (thinking about an `enabled` field instead), but probably we can use it like this, too.

dk1844 (commented):
Nice work overall, found a few items yet to improve.

Also, please add the trailing line break where missing
trailing-line-break-missing

service/src/main/scala/za/co/absa/loginsvc/rest/AuthManagerConfig.scala
TheLydonKing: Used the ConfigValidException. Makes sense to use since it's config related

service/src/main/scala/za/co/absa/loginsvc/rest/AuthManagerConfig.scala
TheLydonKing: Implemented as suggested. Thanks for the tip.

service/src/main/scala/za/co/absa/loginsvc/rest/AuthManagerConfig.scala
TheLydonKing: Ah, this is leftover from a previous implementation. Missed it.

I've removed it.

README.md
TheLydonKing: My Worry is giving the user too many properties to fiddle with to get this working.

So thought it would be a good idea to collapse the functionality into one property.

jakipatryk: Why is removing the (sub)config of a given provider not enough? Do we need two options to have some method disabled?

TheLydonKing: I think for the sake of temporarily disabling methods, whether for debugging or if the auth endpoint goes down for some reason for example,, it's nice to be able to disable it without having to remove everything from the config. Otherwise the user would have to cut and paste the config or comment everything out so they don't lose that information.

PR #55, opened 2023-07-20
Feature/#44 config using native scala types

TheLydonKing:
- Implemented Pureconfig configprovider in order to read Yaml config file into more scala friendly classes and types
- Changed and implemented tests to utilize pureconfig


service/src/main/scala/za/co/absa/loginsvc/rest/config/ConfigProvider.scala
```diff
@@ -0,0 +1,68 @@
...
+import pureconfig._
+import pureconfig.generic.auto._
+import pureconfig.module.yaml._
+import za.co.absa.loginsvc.rest.config.actuator.GitConfig
+import za.co.absa.loginsvc.rest.config.auth._
+import za.co.absa.loginsvc.rest.config.validation.ConfigValidationException
+import java.nio.file.Files
+
+@Component
+class ConfigProvider(@Value("service\\src\\main\\resources\\application.yaml") yaml : String, @Value("false") testEnvironment: Boolean) {
```
jakipatryk, line 30: Windows path?

service/src/main/scala/za/co/absa/loginsvc/rest/config/auth/UsersConfig.scala
```diff
@@ -56,22 +56,16 @@ case class UsersConfig(knownUsers: Array[UserConfig], order: Int)
...
-  def init(): Unit = {
-    this.validate().throwOnErrors()
-  }
 }
 
-@ConstructorBinding
 case class UserConfig(username: String,
                        password: String,
-                       email: String, // may be null
+                       email: Option[String], // may be null
```
dk1844, line 61:
```suggestion
                       email: Option[String],
```
Seeing this become an `Option` makes me happy. What about the comment - I assume the value will be `None`, not null when not found.

service/src/main/scala/za/co/absa/loginsvc/rest/service/JWTService.scala
```diff
@@ -21,16 +21,16 @@ import io.jsonwebtoken.{Jwts, SignatureAlgorithm}
...
-
+import za.co.absa.loginsvc.rest.config.ConfigProvider
 import java.security.{KeyPair, PublicKey}
 import java.time.Instant
 import java.time.temporal.ChronoUnit
 import java.util.Date
 
 @Service
-class JWTService @Autowired()(jwtConfig: JwtConfig) {
+class JWTService @Autowired()(configProvider: ConfigProvider) {
```
dk1844, line 31: This feels a bit weird, because it is only the JTW config this service should have access to (analogically elsewhere).

I think I understand why you did it this way - ConfigProvider as a `@Component` can be autowired while the individual configs cannot. However, perhaps there is a middle ground - if the type was `JWTConfigProvider` (a trait that regular `ConfigProvider` implements and only has jwt-config-related method) that would hide the other config methods here -- and for the future -- it would pose no assumptions if the config @component should be one or more.

You do not need to follow this to the point, I am trying to explain what seemed a bit off with some background.

service/src/main/scala/za/co/absa/loginsvc/rest/config/ConfigProvider.scala
TheLydonKing: NeverMind, I misunderstood.

It's fixed now. Thanks for pointing this out.

service/src/main/scala/za/co/absa/loginsvc/rest/config/provider/ConfigProvider.scala
```diff
@@ -0,0 +1,96 @@
...
+import pureconfig.module.yaml._
+import za.co.absa.loginsvc.rest.config.actuator.GitConfig
+import za.co.absa.loginsvc.rest.config.auth._
+import za.co.absa.loginsvc.rest.config.validation.ConfigValidationException
+import za.co.absa.loginsvc.rest.config.{BaseConfig, JwtConfig}
+
+import java.nio.file.{InvalidPathException, Paths}
+
+@Component
+class ConfigProvider(@Value("${DefaultYamlPath:service/src/main/resources/application.yaml}") yamlContent: String)
```
dk1844, line 32: So many `var`s in this class. Is there any reason why they all couldn't be `val`s?

service/src/main/scala/za/co/absa/loginsvc/rest/config/provider/ConfigProvider.scala
```diff
@@ -0,0 +1,96 @@
...
+  extends JwtConfigProvider
+    with AuthConfigProvider {
+
+  private var yamlConfig: YamlConfigSource = {
+    if (isValidPath(yamlContent))
+      YamlConfigSource.file(yamlContent)
+    else
+      YamlConfigSource.string(yamlContent)
```
dk1844, line 40: What is the point of this? Why do expect, in the case of the path being invalid, to find a usable yaml definition in the very same string?

service/src/main/scala/za/co/absa/loginsvc/rest/config/provider/ConfigProvider.scala
dk1844, line 32: Also, how do you handle config overrides?
In the previous version, one could pass `spring.config.location=/path/to/config`. Here, it seems you hardcode a path to `application.yaml` that is packaged inside the war/jar.

We need to be able to build the package without an internal config and pass an external one somehow. Sorry if you have thought about this and I am missing it.

https://pureconfig.github.io/docs/ talks about using `ConfigSource.default` as a way to define an overridable source, then e.g. `-Dconfig.file=path/to/config-file` can be used for externilized config.

... in the end, if Yaml was too hard for us, with pureConfig/tsConfig, we could be using HOCON. That's not a problem, since it also allows using objects in config (properties do not). _This would be much cheaper to do now than later._

dk1844: One more point here: later, in when dockerized, we need to slip the configuration to the running war, too. This we currently do via setting env var `SPRING_CONFIG_ADDITIONAL-LOCATION=/opt/application.yaml`

Bear this in mind that some form of configuration override is probably needed as well. Does pureConfig support Typesafe config's `CONFIG_FILE` env property for application.conf external loading?

service/src/main/scala/za/co/absa/loginsvc/rest/config/provider/ConfigProvider.scala
```diff
@@ -0,0 +1,96 @@
...
+
+  //GitConfig needs to be initialized at startup
+  getGitConfig
+
+  def getBaseConfig : BaseConfig = {
+    createConfigClass[BaseConfig]("loginsvc.rest.config").
+      getOrElse(BaseConfig(""))
+  }
+
+  def getGitConfig : GitConfig = {
```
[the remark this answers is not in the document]
TheLydonKing: Made Private

service/src/main/scala/za/co/absa/loginsvc/rest/config/provider/ConfigProvider.scala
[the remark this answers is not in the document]
TheLydonKing: I've changed this, please see comment below

TheLydonKing: Fixed

service/src/main/scala/za/co/absa/loginsvc/rest/config/provider/ConfigProvider.scala
TheLydonKing: Removed. This was left over from an older implementation

service/src/main/scala/za/co/absa/loginsvc/rest/config/provider/ConfigProvider.scala
TheLydonKing: Ok, so @Value in Spring checks for values matching that key in config files, environment variables and program arguments.

So I've changed it so that it mimics spring.config.location which can be set in either program arguments or environment variables, so this should sort that out. I've added this to the readme.

TheLydonKing:
Apologies for the force push, needed to do so for the rebase to work work correctly

service/src/test/scala/za/co/absa/loginsvc/rest/config/provider/ConfigProviderTest.scala
```diff
@@ -0,0 +1,55 @@
...
+import org.scalatest.matchers.should.Matchers
+import za.co.absa.loginsvc.rest.config.auth.{ActiveDirectoryLDAPConfig, UsersConfig}
+import za.co.absa.loginsvc.rest.config.{BaseConfig, JwtConfig}
+
+class ConfigProviderTest extends AnyFlatSpec with Matchers  {
+
+  private val configProvider : ConfigProvider = new ConfigProvider("service/src/test/resources/application.yaml")
+
+  "The baseConfig properties" should "Match" in {
+    var baseConfig: BaseConfig = configProvider.getBaseConfig
```
dk1844, line 29:
```suggestion
    val baseConfig: BaseConfig = configProvider.getBaseConfig
```
... and 3 below

service/src/main/scala/za/co/absa/loginsvc/rest/config/auth/UsersConfig.scala
dk1844: 😆 what a fix (yes, an option can be None, but that is sort of expected). The comment is not wrong, but my suggestion was to remove it.

TheLydonKing: Ah, will do

service/src/main/scala/za/co/absa/loginsvc/rest/config/provider/ConfigProvider.scala
```diff
@@ -0,0 +1,86 @@
...
+import pureconfig._
+import pureconfig.generic.auto._
+import pureconfig.module.yaml._
+import za.co.absa.loginsvc.rest.config.actuator.GitConfig
+import za.co.absa.loginsvc.rest.config.auth._
+import za.co.absa.loginsvc.rest.config.validation.ConfigValidationException
+import za.co.absa.loginsvc.rest.config.{BaseConfig, JwtConfig}
+
+@Component
+class ConfigProvider(@Value("${spring.config.location}") yamlContent: String)
```
dk1844, line 31: This is a better than bofore. If you have the app launching in your hand directly, you can use this.

When you are e.g. building a Docker image and where want to set env variable, this becomes harder (btw. you cannot set env var containing `.`). I tried to change it to a different ENV var, e.g. `LS_CONFIG_LOCATION`), but that did not work either.

A viable option with this code seems to be update `JAVA_OPTS`, e.g.:
`ENV JAVA_OPTS="$JAVA_OPTS -Dspring.config.location=/opt/application.yaml"`

So it **seems to work** that way, but it is a bit harder to use.

What would help if there was a smarter config loader that would:
 - check the Java env var like now, e.g. `spring.config.location` or perhaps rather `ls.config.location`
 - check actual ENV VAR using `System.getenv("LS_CONFIG_LOCATION")`
- other options, possibly
With these option, there would be a priority ordering of which to use first.

I leave this comment up to you if you feel this should be implemented here or if you wish to create a future enhancement issue for this. As I said, there is a was to make it work for now using the `JAVA_OPTS` apparently from the system env var perspective.

TheLydonKing: Ok, I'd rather just fix it now. I have an idea, let me just test it quick.

service/src/main/resources/example.application.yaml
```diff
@@ -29,6 +29,7 @@ loginsvc:
             - username: "user1"
               password: "password1"
               groups:
+                - "group1"
```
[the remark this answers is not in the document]
TheLydonKing: Let me try and figure this out... I think from the Pureconfig perspective an empty tag is the same as not having the tag there at all so it would have to be an option... but I think I can find a way around it.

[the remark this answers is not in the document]
TheLydonKing: Ok, so it seems for PureConfig to be able to ensure that the field is required but it's contents can be empty, I need to set it like so:
groups: Array[Option[String]]

And the Config should look like this to be empty:
            - username: "user1"
              password: "password1"
              groups:
                -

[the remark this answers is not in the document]
jakipatryk: what about 
```
- username: "user1"
password: "password1"
groups: []
```
? (from top of my head, might not work)

`groups: Array[Option[String]]` looks totally unreasonable to me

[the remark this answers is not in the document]
TheLydonKing: `- username: "user1"
password: "password1"
groups: []`

This works

`groups: Array[Option[String]]` Not sure how to get around this since PureConfig seems to need values to be typed as Options for them to be considered optional.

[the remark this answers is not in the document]
jakipatryk: > - username: "user1" password: "password1" groups: [] .... This works

with `groups: Array[String]`?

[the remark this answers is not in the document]
TheLydonKing: No, with the Option

Currently trying with
`groups: Array[String]`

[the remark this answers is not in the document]
TheLydonKing: It works.

Changing it back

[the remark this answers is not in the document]
TheLydonKing: Changed it back

So this works:

Config:
`- username: "user1"
password: "password1"
groups: []`

Typing:
`groups: Array[String]`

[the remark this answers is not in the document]
dk1844: ⬆️ sounds good to me.

Another observation I forgot to mention - when I tried it last time, the weirdest thing was that there was no error about invalid config at all, the whole auth provider was ignored (refused to authenticate even users with groups etc)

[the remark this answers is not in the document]
TheLydonKing: I think since it's because the auth methods are optional, so if the auth method is unparsable then it takes it as if it wasn't there in the 1st place. I can add logging to indicate if a auth method is being skipped?

service/src/main/scala/za/co/absa/loginsvc/rest/config/provider/ConfigProvider.scala
dk1844: Confirming that `SPRING_CONFIG_LOCATION` env var works for me 😄

service/src/main/resources/example.application.yaml
```diff
@@ -28,7 +28,7 @@ loginsvc:
           known-users:
             - username: "user1"
               password: "password1"
-              groups:
+              groups: []
```
dk1844, line 31: 😀

dk1844 (approved):
### PR Matrix
- [x] code reviewed
- [x] pulled
- [x] built
- [x] run
- [x] tested a bit (some config variations, using the `SPRING_CONFIG_LOCATION` env var)

PR #57, opened 2023-07-21
Feature/56 docker tomcat in-place update

dk1844:
`tomcat.xml` has been removed and the SSL connector is now defined in code of the Dockerfile.
Also, optional `ARG LS_PREFIX` has been added for situations when running in non-default context


service/Dockerfile
```diff
@@ -61,8 +59,8 @@ RUN echo yes | keytool -import -file /opt/ldapcert.pem -alias ldaps -keystore $J
 FROM base AS base-ssl-true
 ENV SSL_ENABLED=true
 RUN echo "This stage sets SSL=$SSL_ENABLED"
-# Copy Tomcat properties (to enable SSL)
-COPY ./src/main/resources/tomcat-ssl/server.xml /usr/local/tomcat/conf/
+# Enable SSL: Add our SSL connector after <Service name="Catalina"> is found in server.xml; .bak = backup
+RUN sed -i.bak '/<Service name="Catalina">/a <Connector port="8443" protocol="HTTP/1.1" SSLEnabled="true" maxThreads="150" scheme="https" secure="true" keystoreFile="/usr/local/tomcat/conf/selfsigned.p12" keystorePass="changeit" clientAuth="false" acceptCount="100"/>' /usr/local/tomcat/conf/server.xml
```
dk1844, line 63: In case you were wondering on how this is done, the method used is this:
Basically, `sed` does not only replace stuff (most common usage) but it can also append things after a match using the `a` command:
`sed '/^anothervalue=.*/a after=me' test.txt` 

[src](https://fabianlee.org/2018/10/28/linux-using-sed-to-insert-lines-before-or-after-a-match/)

PR #63, opened 2023-08-22
JWKS demo

### LINKED ISSUE #62: supporting the JWKS format for the public key in the login service
## Background
supporting the JWKS format for the public key in the login service

## Feature
There are libraries that work with JWT that implement JWKS endpoint consumption for user's simplicity.

https://medium.com/javarevisited/json-web-key-set-jwks-94dc26847a34
https://github.com/jwtk/jjwt#jwk has unreleased support?

This library can do it: https://github.com/felx/nimbus-jose-jwt

Here's the RFC: https://datatracker.ietf.org/doc/html/rfc7517, Chapters 4 and 5, and an example https://datatracker.ietf.org/doc/html/rfc7517#appendix-A.1

Thanks, @kevinwallimann

kevinwallimann:
Closes #62 

This is just a draft to demonstrate the JWKS endpoint for the public key. JWKS is a standard defined in https://datatracker.ietf.org/doc/html/rfc7517 and is implemented e.g. by 
`org.springframework.boot:spring-boot-starter-oauth2-resource-server`. Resource servers can just add this dependency instead of implementing fetching and parsing the endpoint themselves. Instead, it's just one line of application configuration, i.e.

`spring.security.oauth2.resourceserver.jwt.jwk-set-uri=http://localhost:9090/token/public-key-jwks`

and a few lines in the security configuration (which you need for any JWT security)

## Problems
The following issues need to be discussed before merging
- I had to add the Nimbus JOSE JWT library https://bitbucket.org/connect2id/nimbus-jose-jwt/src/master/, https://connect2id.com/products/nimbus-jose-jwt, which covers similar functionality as https://github.com/jwtk/jjwt. Unfortunately, jjwt supports JWK only in master, but not in the latest released version v0.11.5

### DIFF
--- a/service/src/test/scala/za/co/absa/loginsvc/rest/service/JWTServiceTest.scala
+++ b/service/src/test/scala/za/co/absa/loginsvc/rest/service/JWTServiceTest.scala
@@ -16,6 +16,8 @@
 
 package za.co.absa.loginsvc.rest.service
 
+import com.nimbusds.jose.JWSAlgorithm
+import com.nimbusds.jose.jwk.KeyUse
 import io.jsonwebtoken.{Claims, Jws, Jwts}
 import org.scalatest.flatspec.AnyFlatSpec
 import za.co.absa.loginsvc.model.User
@@ -115,4 +117,23 @@ class JWTServiceTest extends AnyFlatSpec {
     assert(actualGroups === userWithGroups.groups)
   }
 
+  behavior of "jwks"
+
+  it should "return a JWK that is equivalent to the `publicKey`" in {
+    import scala.collection.JavaConverters._
+
+    val publicKey = jwtService.publicKey
+    val jwks = jwtService.jwks
+    val rsaKey = jwks.getKeys.asScala.head.toRSAKey
+
+    assert(publicKey == rsaKey.toPublicKey)
+  }
+
+  it should "return a JWK with parameters" in {
+    import scala.collection.JavaConverters._
+
+    val jwk = jwtService.jwks.getKeys.asScala.head
+    assert(jwk.getAlgorithm == JWSAlgorithm.RS256)
+    assert(jwk.getKeyUse == KeyUse.SIGNATURE)
+  }
 }
--- a/service/src/test/scala/za/co/absa/loginsvc/rest/service/JWTServiceTest.scala
+++ b/service/src/test/scala/za/co/absa/loginsvc/rest/service/JWTServiceTest.scala
@@ -91,6 +91,17 @@ class JWTServiceTest extends AnyFlatSpec {
     }
   }
 
+  it should "return a JWT kid" in {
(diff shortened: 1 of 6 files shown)


jakipatryk:
> Instead, it's just one line of application configuration, i.e.
`spring.security.oauth2.resourceserver.jwt.jwk-set-uri=http://localhost:9090/token/public-key-jwks`

Does it support JWKS refresh out of the box? If yes, how often does it do that?

kevinwallimann:
> > Instead, it's just one line of application configuration, i.e.
> > `spring.security.oauth2.resourceserver.jwt.jwk-set-uri=http://localhost:9090/token/public-key-jwks`
> 
> Does it support JWKS refresh out of the box? If yes, how often does it do that?

Excellent question! It is refreshed every 5 mins, see https://bitbucket.org/connect2id/nimbus-jose-jwt/src/9.24.4/src/main/java/com/nimbusds/jose/jwk/source/DefaultJWKSetCache.java

I've just tried it out myself

dk1844:
I looked it over and I agree that it would be useful to have as an addition.
It is a pity that [jjwt](https://github.com/jwtk/jjwt#jwk) do not have it in release right now, based on some of [their information](https://github.com/jwtk/jjwt/issues/780) and the [state of the milestone](https://github.com/jwtk/jjwt/milestone/21), it might be "soon".

I don't mind the Nimbus JOSE JWT library that much, considering we are hopefully looking forward to replacing it eventually.

I have created a small update on @kevinwallimann's work in https://github.com/AbsaOSS/login-service/pull/64

@kevinwallimann, if you feel like finishing this, be my guest; if you don't want to or don't have time, we could do it, I hope.

dk1844 (approved):

service/src/test/scala/za/co/absa/loginsvc/rest/service/JWTServiceTest.scala
```diff
@@ -115,4 +117,23 @@ class JWTServiceTest extends AnyFlatSpec {
...
+    val jwks = jwtService.jwks
+    val rsaKey = jwks.getKeys.asScala.head.toRSAKey
+
+    assert(publicKey == rsaKey.toPublicKey)
+  }
+
+  it should "return a JWK with parameters" in {
+    import scala.collection.JavaConverters._
+
+    val jwk = jwtService.jwks.getKeys.asScala.head
```
dk1844, line 135:
```suggestion
    val keys = jwtService.jwks.getKeys.asScala
    assert(keys.length == 1, "One JWK is expected to be generated now")

    val jwk = keys.head
```

Perhaps we could assert that there is only 1 key now to be checked. Should the set get extended at some point, it would be nice for test to fail to force its adjustment 😈

dk1844 (approved):
LGTM, checked out, ran tests this time.

PR #64, opened 2023-08-22
Feature/jwks kid

dk1844:
Adding `kid` to JWT claims and to JWKS + a simple unit test added.

### DIFF
--- a/service/src/main/scala/za/co/absa/loginsvc/rest/service/JWTService.scala
+++ b/service/src/main/scala/za/co/absa/loginsvc/rest/service/JWTService.scala
@@ -38,6 +38,7 @@ class JWTService @Autowired()(jwtConfigProvider: JwtConfigProvider) {
 
   private val jwtConfig = jwtConfigProvider.getJWTConfig
   private val rsaKeyPair: KeyPair = Keys.keyPairFor(SignatureAlgorithm.valueOf(jwtConfig.algName))
+  private val kid: String = s"rsa-${jwtConfig.algName}-1"
 
   def generateToken(user: User): String = {
     import scala.collection.JavaConverters._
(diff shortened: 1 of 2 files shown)


service/src/main/scala/za/co/absa/loginsvc/rest/service/JWTService.scala
```diff
@@ -38,6 +38,7 @@ class JWTService @Autowired()(jwtConfigProvider: JwtConfigProvider) {
 
   private val jwtConfig = jwtConfigProvider.getJWTConfig
   private val rsaKeyPair: KeyPair = Keys.keyPairFor(SignatureAlgorithm.valueOf(jwtConfig.algName))
+  private val kid: String = s"rsa-${jwtConfig.algName}-1"
```
kevinwallimann, line 41: The RFC https://datatracker.ietf.org/doc/html/rfc7517#section-4.5 states that "The "kid" (key ID) parameter is used to match a specific key"

So I think that it should be the thumbprint of the JWT. For example, when you change the keypair, I think the kid should also change.

dk1844: Thanks. I read the _The "kid" (key ID) parameter is used to match a specific key_, but my understanding of that sentence was that when we had multiple keys in the jwks at the same time, we should make distinguish them by `kid`s (that would have worked with a 'fixed' kid as I did it. But for rotation, it would not, you are right. 

I have changed the implementation as suggested.

kevinwallimann: I also understood it that way initially, but I've seen code (an earlier version of that DefaultJWKSetCache btw) that used the kid to determine whether the cache should be evicted or not. And I can't really argue that "specific key" only means a certain key within the list, but not a key A vs a key B.

kevinwallimann (approved):

PR #65, opened 2023-08-23
jwks: kid changed to be a public rsakey thumbprint 2

dk1844:
@kevinwallimann

PR #66, opened 2023-08-23
remove static val kid

kevinwallimann:


dk1844 (approved):
Yup, that's better than #65

PR #67, opened 2023-08-24
#42 `/token/generate` now has optional param `groups-prefix`

### LINKED ISSUE #42: Groups filtering
## Background
It may happen, that a user's list of groups is too long. Both in practicality and in terms of token being too large to save in limited spaces (4K limit for cookies for example).

For this reason, it may be useful to request user token with an optional filtering field that will limit the groups to the set of interested values.

## Feature
Provide `/token/generate` rest call (either introduce an optional param there or create another endpoint) where groups filter can be employed.

Consider the practical options for filtering. 
**From the practical standpoint, prefix filtering should be enough.**

## Example [Optional]
`/token/generate?group-prefix=abcd-`
`/token/generate?group-prefix=abcd-|efgh-|something-else-`
(`/token/generate?group-suffix=-admin`)
(`/token/generate?group-contains=-ursa-unify-`)
(`/token/generate?group-regex=\d{4}-\d\d-\d\d`)

(just ideas, not all of these need to be implemented if not needed)

dk1844:
Endpoint  `POST /token/generate` now has optional param `groups-prefix`.

e.g. `POST {{ls_host}}/token/generate?groups-prefix=pam`

- Open API updated - swagger rendering the query param info:
generate-token-prefix

Closes #42


TheLydonKing (approved):
LGTM

Makes sense

PR #71, opened 2023-08-29
Use / as server url

### LINKED ISSUE #59: Rest Methods returning errors on Swagger
## Describe the bug
The Rest Methods seem to be working incorrectly when used through the Swagger page on both the DEV and UAT deployments:
- https://login-service-dev.npintdebdtools.aws.dsarena.com/swagger-ui/index.html
- https://login-service-uat.npintdebdtools.aws.dsarena.com/swagger-ui/index.html

## Possible Reason:
We can see that the URL used in the Swagger and it's methods is HTTP instead of HTTPS. This may be the cause of the error and should be the starting point in investigation and rectification.

kevinwallimann:
Closes #59 

The approach in https://github.com/absa-group/ingestionaas-application/issues/128 didn't work, only an empty string was shown in the url field.

I followed this approach:

https://github.com/springdoc/springdoc-openapi/issues/726

The url is now shown as "/", but the request works


dk1844 (approved):
Thank you, @kevinwallimann, for the contribution.

LGTM, tested on dev, works as described.

ls-swagger-slash

PR #73, opened 2023-09-06
group-prefixes - multiple values at once

### LINKED ISSUE #42: Groups filtering

dk1844:
This PR enhances work done in #67, now there are multiple user group prefixes to be requested at once.

Original: 
 - `POST /token/generate?groups-prefix=just-one-prefix` 
Newly:
 - `POST /token/generate?group-prefixes=prefix1,prefix2,prefix3`.

Added a few tests.

Closes #42


jakipatryk (approved):
LGTM (just checked the code)

PR #78, opened 2023-10-09
Feature/75 refresh token simple extend

### LINKED ISSUE #75: Refresh token implementation
## Background
Current token validity is set in config in hours (4). For convenient usage, we would need a full day (9h), but this may be too much of a risk when being stolen.

## Feature
We need to create a refresh token mechanism that will last about a working day (~9h) to be used for requesting a regular JWT auth token (whose validity time will shorten to about 15 min).

Solution Ideas:
1. Separate endpoints for JWT acquiring - one with for credentials supply (to obtain auth JWT and refresh JWT), another for refresh of auth JWT only (supply still valid refresh JWT, obtain new auth JWT)
2. another similar mechanism

Consider researching/confirming what is the norm.

Use a secure cookie for the refresh token as in: https://www.geeksforgeeks.org/jwt-authentication-with-refresh-tokens/ (recommended the cookie to be `httpOnly=true` and `secure=true`) ?

The refresh token should contain `username` to be only usable for a particular user.

Phase 2: #76

dk1844:
This PR brings the ability to refresh access tokens as outlined in #75. Specifically:

## `/token/generate` also yields a refresh token in Cookie
Previously, `/token/generate` only returned a single (access, but it was not marked as such) token. Now, the access token is still returned in the body
```json
{
    "token": "..."
}
```
and the refrehs token is set to a Cookie: `refresh=...`. Otherwise unchanged, except that a new field `type=access|refresh` has been added to signify the token type. E.g.:
```json
{
  "sub": "ab01234",
  "exp": 1696941933,
  "iat": 1696941903,
  "kid": "aaaaaaaaaaaaaaaaaaaaa-bbbbbbbbbbbbbbbbbbbbb",
  "groups": [
    "MyAwesomeGroup1",
    "AdminOfAllUniverse"
  ],
  "email": "Daniel.Kavan@absa.africa",
  "displayname": "Daniel Kavan (CZ)",
  "type": "access"
}
```

The refresh token only contains `sub=username` and `type=refresh` (apart from the needed `iat`/`exp`), e.g.
```json
{
  "sub": "ab01234",
  "exp": 1696941564,
  "iat": 1696941504,
  "type": "refresh"
}
```

## `/token/refresh` added to get a refreshed `access` token

An endpoint `/token/refresh` has been added to refresh the current access token via an issued refresh token.
You can send exactly what you received from `/token/generate` before your `refresh` token expires (access may have expired already) - i.e. access token in payload, refresh token in Cookie.

refresh-access-token-cookie

Closes #75


jakipatryk:
Shouldn't the refresh token be returned as HTTP-only cookie?

dk1844:
> Shouldn't the refresh token be returned as HTTP-only cookie?

A `secure=true` and `httpOnly=true` cookie is an option. This is useful if used by browsers.
However, on the other hand, there may be situations where the cookie could prove problematic - clients may need special care to follow those, or I think I read about other situations ... let me look it up ... basically, this is more approachable for programmatic clients - simply a payload-based solution.

fateeand:
> > Shouldn't the refresh token be returned as HTTP-only cookie?
> 
> A `secure=true` and `httpOnly=true` cookie is an option. This is useful if used by browsers. However, on the other hand, there may be situations where the cookie could prove problematic - clients may need special care to follow those, or I think I read about other situations ... basically, this is more approachable for programmatic clients - simply a payload-based solution.
> 
> The FE client can save the refresh token to local/session storage.
> 
> If needed, I can rewrite it to secure,http-only cookie.
> 
> //edit: I remember now: if you set the refresh token to cookie, I think is unnecessarily sent with every request. But you don't need that, you only need it to be used when you are actually refreshing. (this may even be unwanted from security perspective -- this may increase the chance of the refresh token being stolen?
> 
> We could ask @fateeand, too. @fateeand what is your preference or can you work with both (refresh token being in the payload to be saved to localStorage -- I am sure that this you can work with. What about the secure,httpOnly cookie?)

Since the refresh token rotation is implemented (`/token/refresh` returns new refresh token, right?) we can store it in local storage. In this case refresh token lifetime is also short and linked to access token lifetime.

Read this article for reference
https://auth0.com/blog/refresh-tokens-what-are-they-and-when-to-use-them/#You-Can-Store-Refresh-Token-In-Local-Storage

dk1844:
> Since the refresh token rotation is implemented (`/token/refresh` returns new refresh token, right?) we can store it in local storage. In this case refresh token lifetime is also short and linked to access token lifetime.
> 
> Read this article for reference https://auth0.com/blog/refresh-tokens-what-are-they-and-when-to-use-them/#You-Can-Store-Refresh-Token-In-Local-Storage

Currently, the service does not return a new refresh token on each `/token/refresh`, but rather behave as if it does, because we may implement this in the future to limit possible misuse.

But its validity will only < working day, so it is still fine.

fateeand:
> > Since the refresh token rotation is implemented (`/token/refresh` returns new refresh token, right?) we can store it in local storage. In this case refresh token lifetime is also short and linked to access token lifetime.
> > Read this article for reference https://auth0.com/blog/refresh-tokens-what-are-they-and-when-to-use-them/#You-Can-Store-Refresh-Token-In-Local-Storage
> 
> Currently, the service does not return a new refresh token on each `/token/refresh`, but rather behave as if it does, because we may implement this in the future to limit possible misuse.
> 
> But its validity will only < working day, so it is still fine.

Ok, so for now this refresh token doesn't have much sense from the security point of view.
If the attacker will manage to obtain it once, then he can get new access tokens within the working day, basically same story as having long-living access token.
At least we won't force users to relogin and a bit more secure anyways. I will keep frontend-based user inactivity tracking functionality as well for now.

So looking forward to a refresh token rotation mechanism, and we should be safe.

service/src/main/scala/za/co/absa/loginsvc/rest/RestResponseEntityExceptionHandler.scala
```diff
@@ -0,0 +1,55 @@
...
+    logger.error(exception.getMessage)
+    ResponseEntity.badRequest().body(RestMessage(exception.getMessage))
+  }
+
+  @ExceptionHandler(value = Array(classOf[IllegalArgumentException]))
+  def handleIllegalArgumentException(exception: IllegalArgumentException): ResponseEntity[RestMessage] = {
+    ResponseEntity.badRequest().body(RestMessage(exception.getMessage))
+  }
+
+}
```
TheLydonKing, line 55: line break

TheLydonKing (approved):
Went through all the code, looks good to me.

dk1844:
> Shouldn't the refresh token be returned as HTTP-only cookie?

Redone as a secure HTTP-only cookie. Tests updated, swagger updated. Rebased to the lastest master.

README.md
```diff
@@ -124,6 +161,7 @@ loginsvc:
         private-key-field-name: "privateKey"
         public-key-field-name: "publicKey"
         access-exp-time: 15min
+        key-rotation-time: 9h
```
TheLydonKing, line 164: This should be refresh-exp-time not key-rotation time.

In AWS Secrets mode, we don't rotate the key within the app, the scheduler is used to poll aws secrets manager for changes in the secret.
```suggestion
         access-exp-time: 15min
         refresh-exp-time: 9h
         poll-time: 30min
```

TheLydonKing (approved):
Read through code and only 1 minor comment on Readme, besides that I tested it and was able to get an access key with a refresh key cookie and also all tests passed on `sbt testonly`

fateeand:
Please don't forget to set access-exp-time to 9 hours before merging (a temporary measure until UI support becomes available both in Unify and Aqueduct)

TheLydonKing:
> Please don't forget to set access-exp-time to 9 hours before merging (a temporary measure until UI support becomes available both in Unify and Aqueduct)

I'll do that on the deployment side with the latest Terraform additions

README.md
dk1844: Done in https://github.com/AbsaOSS/login-service/pull/78/commits/a99fcaaa68b3883208d41e7ca11259fcc12dccee, hopefully as you meant it (sorry, the suggestion box seems a bit confusing)

PR #79, opened 2023-11-13
Feature/75 refresh token simple extend

### LINKED ISSUE #75: Refresh token implementation

dk1844:
This PR changes #78 back to a payload-based implementation, implements #75.75. Specifically:

## `/token/generate` also yields a refresh token
Previously, `/token/generate` only returned a single (access, but it was not marked as such) token. Now, the access token & refresh token are the body
```json
{
    "token": "...",
    "refresh": "..."
}
```
Otherwise unchanged, except that a new field `type=access|refresh` has been added to signify the token type. E.g.:
```json
{
  "sub": "ab01234",
  "exp": 1696941933,
  "iat": 1696941903,
  "kid": "aaaaaaaaaaaaaaaaaaaaa-bbbbbbbbbbbbbbbbbbbbb",
  "groups": [
    "MyAwesomeGroup1",
    "AdminOfAllUniverse"
  ],
  "email": "Daniel.Kavan@absa.africa",
  "displayname": "Daniel Kavan (CZ)",
  "type": "access"
}
```

The refresh token only contains `sub=username` and `type=refresh` (apart from the needed `iat`/`exp`), e.g.
```json
{
  "sub": "ab01234",
  "exp": 1696941564,
  "iat": 1696941504,
  "type": "refresh"
}
```

## `/token/refresh` added to get a refreshed `access` token

An endpoint `/token/refresh` has been added to refresh the current access token via an issued refresh token.
You can send exactly what you received from `/token/generate` before your `refresh` token expires (access may have expired already)

ls-token-refresh

Closes #75


jakipatryk (approved):
LGTM (just checked code).

Issue #80, opened 2023-11-21
Externalize config from the image


dk1844:
## Background
Currently, the configuration is backed into the image. This requires having multiple ECR repositories per config. Config could be supplied externally and there should be only one the image per version

## Feature
Find a way to supply the config externally to the image and change the image-building process to support this. Any deployment worklows based needs to be changed, too.

## Proposed Solution [Optional]
Solution Ideas:
Load config from an external path that resides on a mounted volume
1. Require full config there, or
2. Require just override from a builtin default (this can be useful, but dangerous)
3. Require just override from a builtin default, but make sure that built-in misses some fields and thus fails if config is not provided

dk1844:
Closed via #88

PR #88, opened 2024-01-12
#80 `application.conf` is no longer baked into the docker image

dk1844:
`application.conf` is no longer baked into the docker image, but loaded from `/opt/config/application.yaml` (not part of the image, needs to be mounted when run, e.g. via `docker run -v`)

certs copy limited to *.pem (only `pem`s are used anyway)


TheLydonKing (approved):
Read Code, LGTM

PR #91, opened 2024-01-18
#3 add release process support

TheLydonKing:
Added new Workflow that publishes Client Library to Sonatype. This utilizes ci-release dependency. Passwords and Access Keys for the github actions are stored in Secrets.

I have not used Github actions before, so please let me know if the workflow is correct or if their are further changes needed.

### DIFF
--- a/.github/workflows/publish_lib.yml
+++ b/.github/workflows/publish_lib.yml
@@ -16,9 +16,8 @@
 
 name: Release - publish client library
 on:
-  push:
-    branches: [master]
-    tags: ["*"]
+  release:
+    types: [released]
 
 jobs:
   publish:
--- a/.github/workflows/publish_lib.yml
+++ b/.github/workflows/publish_lib.yml
@@ -14,10 +14,11 @@
 # limitations under the License.
 #
 
-name: Release - publish artifacts
+name: Release - publish client library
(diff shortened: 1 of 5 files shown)


.github/workflows/publish_lib.yml
```diff
@@ -0,0 +1,39 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+name: Release - publish client library
+on:
+  push:
+    branches: [master]
+    tags: ["*"]
```
dk1844, line 21:
```suggestion
on:
  push:
    branches:
      - 'master'
      - 'support/**'
    tags: 
      - 'v[0-9]+.[0-9]+.[0-9]+*'
```

We may need to release hotfixes that do not get merged to `master` and will reside in the a support branch.

TheLydonKing: Ok, I think the release version works so we don't create new versions on every push to master as we may be doing changes that don't affect the library at all, so a new version getting pushed to Maven would be pointless.

dk1844 (approved):
LGTM (just read the code)

PR #92, opened 2024-01-22
Mvn central version badge

dk1844:
client library scala 2.12 badge


dk1844:
The badge is already present in the client library readme page https://github.com/AbsaOSS/login-service/blob/master/clientLibrary/README.md

PR #97, opened 2024-02-01
#86 make ldap fields mapping used configurable

### LINKED ISSUE #86: Make LDAP fields mapping used configurable
## Background
Currently, we hardcode fields we found in LDAP (`displayname`, `groups`, etc.). However, this is specific to our LDAP setup and should be generalized via config

## Feature
A description of the requested feature.

## Example [Optional]
A simple example if applicable.

Solution Ideas:
1. 
2. 
3.

TheLydonKing:
Added a configurable field to add additional fields to the JWT claims and to retrieve more information from Ldap.

Please let me know if this implementation makes sense.

Closes #86

### DIFF
--- a/api/src/main/scala/za/co/absa/loginsvc/rest/config/auth/ActiveDirectoryLDAPConfig.scala
+++ b/api/src/main/scala/za/co/absa/loginsvc/rest/config/auth/ActiveDirectoryLDAPConfig.scala
@@ -27,7 +27,7 @@ import za.co.absa.loginsvc.rest.config.validation.{ConfigValidatable, ConfigVali
  * @param url URL to AD LDAP, ex. "ldaps://some.domain.com:636/"
  * @param searchFilter LDAP filter used when searching for groups, ex. "(samaccountname={1})"
  */
-case class ActiveDirectoryLDAPConfig(domain: String, url: String, searchFilter: String, order: Int)
+case class ActiveDirectoryLDAPConfig(domain: String, url: String, searchFilter: String, order: Int, attributes: Array[String])
   extends ConfigValidatable with DynamicAuthOrder
 {
 
--- a/api/src/main/scala/za/co/absa/loginsvc/rest/config/auth/ActiveDirectoryLDAPConfig.scala
+++ b/api/src/main/scala/za/co/absa/loginsvc/rest/config/auth/ActiveDirectoryLDAPConfig.scala
@@ -27,7 +27,7 @@ import za.co.absa.loginsvc.rest.config.validation.{ConfigValidatable, ConfigVali
  * @param url URL to AD LDAP, ex. "ldaps://some.domain.com:636/"
  * @param searchFilter LDAP filter used when searching for groups, ex. "(samaccountname={1})"
  */
-case class ActiveDirectoryLDAPConfig(domain: String, url: String, searchFilter: String, order: Int, attributes: Array[String])
+case class ActiveDirectoryLDAPConfig(domain: String, url: String, searchFilter: String, order: Int, attributes: Option[Array[String]])
   extends ConfigValidatable with DynamicAuthOrder
 {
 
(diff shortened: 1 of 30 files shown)


TheLydonKing:
Using Map[String, Option[AnyRef]] for the new field may be a bit overkill but logically it makes sense to me.

Let me know if you disagree or have a better solution.

api/src/main/scala/za/co/absa/loginsvc/rest/service/JWTService.scala
```diff
@@ -76,8 +76,11 @@ class JWTService @Autowired()(jwtConfigProvider: JwtConfigProvider) {
       .setIssuedAt(issuedAt)
       .claim("kid", publicKeyThumbprint)
       .claim("groups", groupsClaim)
-      .applyIfDefined(user.email)((builder, value: String) => builder.claim("email", value))
-      .applyIfDefined(user.displayName)((builder, value: String) => builder.claim("displayname", value))
+      .addClaims(
+        user.optionalAttributes.collect {
+          case (key, Some(value)) => key -> value
+        }.asJava
+      )
```
dk1844, line 83: great, this part is key - there should be no change to the current expected fields 👍

api/src/main/scala/za/co/absa/loginsvc/rest/config/auth/ActiveDirectoryLDAPConfig.scala
```diff
@@ -27,7 +27,7 @@ import za.co.absa.loginsvc.rest.config.validation.{ConfigValidatable, ConfigVali
  * @param url URL to AD LDAP, ex. "ldaps://some.domain.com:636/"
  * @param searchFilter LDAP filter used when searching for groups, ex. "(samaccountname={1})"
  */
-case class ActiveDirectoryLDAPConfig(domain: String, url: String, searchFilter: String, order: Int)
+case class ActiveDirectoryLDAPConfig(domain: String, url: String, searchFilter: String, order: Int, attributes: Option[Array[String]])
```
dk1844, line 30: Is this necessary or is there a way to reasonably load no attributes as an empty array?

```suggestion
case class ActiveDirectoryLDAPConfig(domain: String, url: String, searchFilter: String, order: Int, attributes: Array[String])
```

clientLibrary/src/test/scala/za/co/absa/loginclient/authorization/ClaimsParserTest.scala
```diff
@@ -80,23 +81,23 @@ class AccessClaimsParserTest extends AnyFlatSpec with Matchers {
...
+    val displayName = AccessTokenClaimsParser.getOptionalClaims(accessJwt)
+    assert(displayName("displayname").toString.equals(FakeTokens.displayName))
   }
 
   it should "return an empty display name if displayname claim is not present at all" in {
-    val dn = AccessTokenClaimsParser.getDisplayName(missingAllClaimsButSubjectJwt) // does not throw
-    dn shouldBe None
+    val dn = AccessTokenClaimsParser.getOptionalClaims(missingAllClaimsButSubjectJwt) // does not throw
+    dn shouldBe Map.empty
   }
```
dk1844, line 101: These test cases are a bit weird.

`val email = AccessTokenClaimsParser.getOptionalClaims(accessJwt)` does not make sense (it is the optional claims map that get returned


Perhaps you could just either:
(a) merge these test cases for situations where
 - optional claims are present
 - optional claims are not present
  (because now it is the same optional claims map that gets returned)
or
(b) use `getClaim(jwt: Jwt, claimKey: String)` for a specific claim existence test in the current test cases structure. 

Or both somehow.

dk1844 (approved):
LGTM (I tested it locally)

It's great that the JWT structure remained unchanged.
Not sure about `ActiveDirectoryLDAPConfig(..., attributes: Option[Array[String]])` as described below.

api/src/main/scala/za/co/absa/loginsvc/rest/config/auth/ActiveDirectoryLDAPConfig.scala
TheLydonKing: So for PureConfig, if you want the entire tag to be optional then it needs to be an option. Otherwise, we can get rid of the Option but then the tag is mandatory in the config and an empty array has to be supplied if you're not using it. So I guess it's a toss-up between having the field be mandatory in the config or not?

Issue #76, opened 2023-09-26
Refresh token improvement - phase 2


dk1844:
## Background
#75 Introduces a refresh simple token mechanism. This can be improved upon.

## Feature
Simple version (#75) reissues access JWT based on refresh JWT as-is (except for the validity window being anew). But it does not check revalidate with auth backed in the information is still accurate (e.g. against LDAP if user is still there, if has still the same groups, etc)

## Proposed Solution [Optional]
Solution Ideas:
1. Implement user info reload mechanism and use it when refreshing access JWT token - e.g. for LDAP, an AD service user, user information retrieval layer must be added to check no change has happened in the meantime.

Some other ideas in Based on https://www.geeksforgeeks.org/jwt-authentication-with-refresh-tokens/ - considerations around the refresh token being regenerated each time and not being reused?

PR #101, opened 2024-02-26
#76: refresh token improvement phase 2

TheLydonKing:
Branch to improve refresh token functionality.

Now when user logs in via refresh token, we check our auth authorities if the user still exists and if their details are still valid rather than simply copying the details from the old token.

This Pull request will stay in draft form until the following work is complete:
1) Our new Service Account for the login service is made available 
2) Terraform changes to our deployment are done so that service account credentials can be stored in AWS Secrets Manager
3) Config Changes on our deployment environments have been made

Closes: https://github.com/AbsaOSS/login-service/issues/76


api/src/main/scala/za/co/absa/loginsvc/rest/config/auth/ActiveDirectoryLDAPConfig.scala
```diff
@@ -57,3 +72,110 @@ case class ActiveDirectoryLDAPConfig(domain: String, url: String, searchFilter:
...
+      throw ConfigValidationException("Both integratedLdapUserConfig and awsSecretsLdapUserConfig exist. Please choose only one.")
+
+    case (None, None) =>
+      throw ConfigValidationException("Neither integratedLdapUserConfig nor awsSecretsLdapUserConfig exists. One of them should exist.")
+
+    case _ =>
+      val ldapConfig: LdapUser = inConfigAccount.orElse(awsSecretsManagerAccount).get
+      ldapConfig.throwErrors()
+
+      ldapConfig
```
dk1844, line 91:
```suggestion
    case (inConfig@Some(_), None) | (None, awsConfig@Some(_)) =>
      val ldapConfig: LdapUser = inConfig.orElse(awsConfig).get
      ldapConfig.throwOnErrors()
```

Maybe make use of the matching, callers could still pass e.g. `null`s there.

api/src/main/scala/za/co/absa/loginsvc/rest/config/auth/ActiveDirectoryLDAPConfig.scala
```diff
@@ -57,3 +72,110 @@ case class ActiveDirectoryLDAPConfig(domain: String, url: String, searchFilter:
...
+  }
+
+  val username: String = String.format(accountPattern, ldapUserDetails.username)
+  val password: String = ldapUserDetails.password
+}
+
+case class IntegratedLdapUserConfig(username: String, password: String) extends LdapUser
```
dk1844, line 98: How about just `LdapUserCredetialsConfig` or similar - the pair is essentially just _credentials_.

api/src/main/scala/za/co/absa/loginsvc/rest/config/provider/ConfigProvider.scala
```diff
@@ -74,7 +74,10 @@ class ConfigProvider(@Value("${spring.config.location}") yamlPath: String)
         ldapConfig.throwErrors()
         ldapConfig
       }
-    else ActiveDirectoryLDAPConfig("", "", "", 0, None)
+    else{
+      val emptyServiceAccount = ServiceAccountConfig("", Option(IntegratedLdapUserConfig("","")), None)
+      ActiveDirectoryLDAPConfig("", "", "", 0, emptyServiceAccount, None)
+    }
```
dk1844, line 80: This whole method seems a bit weird and more even so with defining an `emptyServiceAccount` here.

https://github.com/AbsaOSS/login-service/pull/101/files#diff-c1c2d4980084933489c41a4356ad8db2c69e005ccddf0fddf877e1d6c586b2e7R63 Documents that a service account must be given, so we should fail if not present -- I see no reason to create an `emptyServiceAccount`.

Also, on a more general note, I wonder - if the whole
```scala
trait AuthConfigProvider {
  def getLdapConfig: ActiveDirectoryLDAPConfig // todo Option?
  def getUsersConfig: UsersConfig // todo Option?
}
```
should not rather yield Option for both configs -- they are, IIRC, optional in their nature.

api/src/main/scala/za/co/absa/loginsvc/rest/service/AuthSearchService.scala
```diff
@@ -0,0 +1,139 @@
...
+  private def searchUsersConfig(username: String): Option[User] = {
+    usersConfig.knownUsersMap.get(username).flatMap { userConfig =>
+      val optionalAttributes: Map[String, Option[AnyRef]] = userConfig.attributes.getOrElse(Map.empty).map {
+        case (k, v) => (k, Some(v))
+      }
+      Some(User(username, userConfig.groups.toList, optionalAttributes))
+    }
+  }
+
+  private def searchLDAP(username: String): Option[User] = {
```
dk1844, line 68: I suggest that these two implementations (searching in built-in users/in LDAP) exist in two separate classes that both implement the same trait, e.g.
```scala
trait UserSearchable {
  def searchForUser: Option[User] 
}
```

api/src/main/scala/za/co/absa/loginsvc/rest/service/AuthSearchService.scala
```diff
@@ -0,0 +1,139 @@
...
+      case u: UsersConfig =>
+        val result = searchUsersConfig(username)
+        if (result.isDefined) return result.get
+      case l: ActiveDirectoryLDAPConfig =>
+        val result = searchLDAP(username)
+        if (result.isDefined) return result.get
+      case _ => //ignore for now
+    }
+    throw new NoSuchElementException(s"Value not found in any object.")
+  }
```
dk1844, line 57: I understand that you are trying to short-circut here (not to lookup in LDAP if found in Config). If you redo the two search option into a common trait impls, there are ways to achieve that in a scala-friendly manner without `return`, e.g. using `takeWhile` - https://stackoverflow.com/a/13596556/1773349

api/src/main/scala/za/co/absa/loginsvc/rest/service/AuthSearchService.scala
```diff
@@ -0,0 +1,139 @@
...
+    val context = getDirContext(serviceAccount.username, serviceAccount.password)
+      try {
+        val users = context
+          .search(adLDAPConfig.domain.split("\\.").map(part => s"dc=$part").mkString(","),
+            adLDAPConfig.searchFilter.replace("{1}", username),
+            getSimpleSearchControls)
+          .asScala.filter(filterSearchResults).map(resultToUserEntry).toList
+
+        if (users.nonEmpty) {
+          println("User found in LDAP")
```
dk1844, line 79:
```suggestion
```
remove `println`s in the end or replace with logging. (here, probably remove, seems like a dev/debug residuum)

TheLydonKing: Yeah, sorry, this is left over from debugging

api/src/main/scala/za/co/absa/loginsvc/rest/config/auth/ActiveDirectoryLDAPConfig.scala
dk1844: The original name `integratedLdapUserConfig` has been left in logging string (`ActiveDirectoryLDAPConfig.scala`).

api/src/main/scala/za/co/absa/loginsvc/rest/service/search/ConfigSearchProvider.scala
```diff
@@ -0,0 +1,37 @@
...
+class ConfigSearchProvider(usersConfig: UsersConfig)
+  extends AuthSearchProvider {
+
+  private val logger = LoggerFactory.getLogger(classOf[ConfigSearchProvider])
+  def searchForUser(username: String): Option[User] = {
+    logger.info(s"Searching for user in config: $username")
+    usersConfig.knownUsersMap.get(username).flatMap { userConfig =>
+      val optionalAttributes: Map[String, Option[AnyRef]] = userConfig.attributes.getOrElse(Map.empty).map {
+        case (k, v) => (k, Some(v))
+      }
```
dk1844, line 32:
```suggestion
val optionalAttributes: Map[String, Option[AnyRef]] = userConfig.attributes.getOrElse(Map.empty).mapValues(Some(_))
```

There is a shorthand if you just want to map values with keys staying the same ⬆️

api/src/main/scala/za/co/absa/loginsvc/rest/AuthManagerConfig.scala
```diff
@@ -33,34 +33,34 @@ import scala.collection.immutable.SortedMap
...
       throw ConfigValidationException("No authentication method enabled in config")
 
-    orderedProviders.foreach(
+    orderedProviders.zipWithIndex.foreach(
       auth => {
-        logger.info(s"Authentication method ${auth._2.getClass.getSimpleName} has been initialized at order ${auth._1}")
-        authenticationManagerBuilder.authenticationProvider(auth._2)
+        logger.info(s"Authentication method ${auth._1.getClass.getSimpleName} has been initialized at order ${auth._2 + 1}")
+        authenticationManagerBuilder.authenticationProvider(auth._1)
       })
```
dk1844, line 55: While technically possible, using `tuple._1` and `tuple._2` is hard to read. The accepted pattern here is to pattern-match to name the fields:

```scala
    orderedProviders.zipWithIndex.foreach { case (authProvider, index) =>
      logger.info(s"Authentication method ${authProvider.getClass.getSimpleName} has been initialized at order ${index + 1}")
      authenticationManagerBuilder.authenticationProvider(authProvider)
    }
```

dk1844 (commented):
I have also noticed that 
`AwsSecretsManagerKeyConfig.keyPair` and `AwsSecretsLdapUserConfig.getUsernameAndPasswordFromSecret` share a fair bit of code in regards to getting value(s) from Secrets manager. Is it worth generalizing into a common helper or not?

api/src/main/scala/za/co/absa/loginsvc/rest/config/auth/ActiveDirectoryLDAPConfig.scala
```diff
@@ -74,30 +74,39 @@ case class ActiveDirectoryLDAPConfig(domain: String,
...
-      ldapConfig.throwErrors()
+    case (inConfig@Some(_), None) =>
+      val ldapConfig: LdapUser = inConfig.get
+      ldapConfig.throwOnErrors()
+
+      ldapConfig
+
+    case (None, awsConfig@Some(_)) =>
+      val ldapConfig: LdapUser = awsConfig.get
+      ldapConfig.throwOnErrors()
```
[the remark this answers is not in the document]
TheLydonKing: So I tried to use the case that you suggested but it kept throwing errors about pattern mismatches or something along the lines of that, so just separated it into 2 separate cases and it worked. I'll have a look at this solution.

api/src/main/scala/za/co/absa/loginsvc/rest/service/search/ConfigSearchProvider.scala
TheLydonKing: Thank you. Will keep this in mind for future.

api/src/main/scala/za/co/absa/loginsvc/rest/AuthManagerConfig.scala
TheLydonKing: Thank you. Will keep this in mind in future.

api/src/main/scala/za/co/absa/loginsvc/rest/config/auth/ActiveDirectoryLDAPConfig.scala
TheLydonKing: Done

api/src/main/scala/za/co/absa/loginsvc/rest/config/auth/ActiveDirectoryLDAPConfig.scala
TheLydonKing: Will fix this now

api/src/main/scala/za/co/absa/loginsvc/rest/service/AuthSearchService.scala
TheLydonKing: So Attempted to use a `takeWhile` `dropWhile` and `find` to accomplish this and while it works, it has the unfortunate side effect of having to run the a search function at least twice in the attempts I tried. So instead I separated the classes like you suggested with a common trait and just simplified this loop to something more readable.

dk1844: How about something like this (trying to do it in a Scala idiomatic fashion without using `return`):

```scala
  def searchUser(username: String): User = {
    val maybeFoundUser = orderedProviders.toIterator // this makes it lazy
      .map(_.searchForUser(username)) // expensive operation done lazily
      .dropWhile(_.isEmpty) // keeps searching until the possibly-found-user is not empty or all providers are exhausted

    if (maybeFoundUser.hasNext) {
      maybeFoundUser.next().getOrElse(throw new IllegalStateException(s"Valid user $maybeFoundUser should be available"))
    } else
      throw new NoSuchElementException(s"No user found by username $username anywhere")
  }
```

Related to this matter, I can see that `AuthSearchServiceTest` is present, but it only has a simple test for this case - `it should "return a matching user" in ...`. I think this class should be enriched to include the checks of multiple auth providers yielding different results:
 - ap1: Some, ap2: throw-should don't be called;
 -  ap1: None, ap2: Some;
 - ap1: None, ap2: None + observe the expected exception.

api/src/main/scala/za/co/absa/loginsvc/rest/config/auth/ActiveDirectoryLDAPConfig.scala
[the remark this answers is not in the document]
TheLydonKing: Here's the error using the suggested patter:
illegal variable in pattern alternative
    case (inConfig@Some(_), None) | (None, awsConfig@Some(_)) =>

api/src/main/scala/za/co/absa/loginsvc/rest/config/auth/ActiveDirectoryLDAPConfig.scala
dk1844: Almost, I see still one left in `throw ConfigValidationException("Neither integratedLdapUserConfig nor awsSecretsLdapUserConfig exists. Exactly one of them should be present.")`

api/src/main/scala/za/co/absa/loginsvc/rest/config/auth/ActiveDirectoryLDAPConfig.scala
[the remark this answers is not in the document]
dk1844: It is fine like you did it now

api/src/main/scala/za/co/absa/loginsvc/rest/service/AuthSearchService.scala
dk1844: I will actually try to do this, I have some other changes in mind - I will create a PR for this.

dk1844: I have created a PR that contains this https://github.com/AbsaOSS/login-service/pull/102

But it has other features - some classes broken down for testability and some example tests added to make the point. Please, give it a look

dk1844: Resolved by #102

dk1844 (approved):
I read it already as Draft 👍

PR #102, opened 2024-03-07
#76 suggestions for refresh token phase 2

dk1844:
rearchitecting suggestions for easier testability of UserSearchService + added some tests

@TheLydonKing


TheLydonKing (approved):
Read through the code and LGTM. Thanks for showing the Iterator. Tried making it lazy with .view and that just broke everything 🥇

PR #103, opened 2024-06-10
100 using spnego similarly like in enceladus to negotiate the auth

### LINKED ISSUE #100: Using SPNEGO similarly like in Enceladus to negotiate the auth

TheLydonKing:
Added Kerberos Auth that can be used:
`curl -i --negotiate -u : "http://localhost:9090/token/generate"`

The Current implementation is kind of working but testing has been difficult to implement. Currently getting a Bad Credentials Exception. Not sure if it's the Ldap search, the Keytab or the kerberos ticket that is currently causing the issue.

closes #100


TheLydonKing:
Excuse the messy code if anyone tries to read this, just trying to find a configuration that works. A lot of code is duplicated or in the wrong folders, this will be cleaned up once we get a working solution.

So Far have tried multiple implementations such as those found in [Enceladus](https://github.com/AbsaOSS/enceladus) as well as the official SPNEGO Documentation and [Examples](https://github.com/spring-projects/spring-security-kerberos/tree/main/spring-security-kerberos-samples)

The current implementation is being called correctly in the security filterchain if "curl -i --negotiate -u : "http://localhost:9090/token/generate" is being called. The Security filterchain is also successfully skipping this auth process if Basic Auth is stipulated in the http message and then BasicAuth is used.

The Current issue that's happening is currently getting a BadCredentials exception when trying to test it using MIT Kerberos on Windows. The Specific Error: `2024-06-12 12:52:52.666  WARN 17980 --- [nio-9090-exec-1] w.a.SpnegoAuthenticationProcessingFilter : Negotiate Header was invalid: Negotiate XXXX

org.springframework.security.authentication.BadCredentialsException: GSSContext name of the context initiator is null`

Current Hypothesis on the issue:

- Something missing from the code. I've been using the above examples to write this code but it's possible that I've misinterpreted something or missed something. Another set of eyes may be able to point this out but currently I'm trying to experiment but not getting any changes.
- Error with Keytab. Keytab was created by Active Directory team but it's possible that a mistake was made or encryption type is incorrect etc.
- KRB5 is incorrect. So far I've tried 2 different KRB5 files and both have given the same output, so not sure if I have the correct data here.
- Testing on windows: Kerberos and keytabs are mostly used on Linux distributions. Attempting to try and use Curl and MIT Kerberos to test this but it's possible that something on Windows is causing issues as kerberos is quite rigid.
- Perhaps a SSL error with verifiying against the Absa KDC specified in the KRB5 files?
- Possibly a cross-domain issue as the test machine is D_ABSA but the Kerberos Keytab and KDC's are CORP domain? Possible that this may be causing issues?

Is anyone able to see a different error that I perhaps missed?

api/src/main/scala/za/co/absa/loginsvc/rest/SecurityConfig.scala
```diff
@@ -49,9 +57,21 @@ class SecurityConfig {
...
+            val kerberos = new KerberosSPNEGOAuthenticationProvider(KerberosConfig)
+            val provider = new RestApiKerberosAuthenticationProvider(KerberosConfig.url, KerberosConfig.searchFilter, KerberosConfig.domain);
+            http.addFilterBefore(
+              RestApiKerberosAuthentication.spnegoAuthenticationProcessingFilter(
+                new ProviderManager(provider, kerberos.kerberosServiceAuthenticationProvider())),
+              classOf[BasicAuthenticationFilter])
+          }
+      }
+
     http.build()
```
dk1844, line 74: I am surprised that it is necessary to force the the `Kerberos...AuthProvider` this way. Isn't this just another `AuthenticationProvider` just like `ConfigUsersAuthenticationProvider` and `ActiveDirectoryLDAPAuthenticationProvider`?

Perhaps it could be provided the same way as those are in `AuthManagerConfig.authManager:47`?

api/src/main/scala/za/co/absa/loginsvc/rest/provider/kerberos/KerberosLdapUserSearch.scala
```diff
@@ -0,0 +1,18 @@
...
+
+class KerberosLdapUserSearch (searchBase: String, searchFilter: String, contextSource: BaseLdapPathContextSource)
+  extends FilterBasedLdapUserSearch(searchBase, searchFilter, contextSource) {
+
+  override def searchForUser(username: String): DirContextOperations = {
+    val user = if (username.contains("@")) {
+      username.split("@").head
+    } else {
+      username
+    }
```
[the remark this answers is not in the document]
dk1844: The same is already at `KerberosSPNEGOAuthentication.scala:98`, too.

api/src/main/scala/za/co/absa/loginsvc/rest/provider/kerberos/KerberosSPNEGOAuthentication.scala
```diff
@@ -0,0 +1,122 @@
...
+    if(groups.isEmpty)
+      AuthorityUtils.NO_AUTHORITIES
+    else {
+      groups.map({group =>
+        val ldapName = new LdapName(group)
+        val role = ldapName.getRdn(ldapName.size()-1).getValue.toString
+        new SimpleGrantedAuthority(role)
+      }).toList.asJava
+    }
+  }
```
dk1844, line 121: What is the point of this piece of code? How does it work?
Users can have bunch of `memberOf` records on `User` node.
Does it extract the the simple name of the member group from e.g. `CN=PAM-PROD-ZA-DEVOPS-USERS,OU=Security,OU=Groups,OU=CORP Accounts,DC=corp,DC=dsarena,DC=com` ? The part that confuses me the most is `ldapName.getRdn(ldapName.size()-1).getValue.toString`

api/src/main/scala/za/co/absa/loginsvc/rest/config/auth/ActiveDirectoryLDAPConfig.scala
```diff
@@ -63,12 +64,32 @@ case class ActiveDirectoryLDAPConfig(domain: String,
...
+        case Some(x) => x.validate()
+        case None => ConfigValidationSuccess
+      }
+      requiredResults.merge(kerberosResults)
     }
     else ConfigValidationSuccess
   }
 }
 
+case class KerberosConfig(krbFileLocation: String, keytabFileLocation: String, debug: Option[Boolean]) extends ConfigValidatable {
```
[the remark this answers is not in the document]
TheLydonKing: Will do. The code is currently just messy from trying to figure out a working solution and debugging.

api/src/main/scala/za/co/absa/loginsvc/rest/SecurityConfig.scala
TheLydonKing: So I'm forcing the Authmanager here so that I can try and debug any issues with the kerberos auth specifically. Just trying to make it easier to find the above issue that is being run into.

api/src/main/scala/za/co/absa/loginsvc/rest/provider/kerberos/KerberosSPNEGOAuthentication.scala
TheLydonKing: Yes, as far as I can see is that this is a way of handling Ldap responses and the code above is how to find the relevant information we need to grant the authority. I was just trying to test this as this how enceladus handles it, so wanted to see if this is a better method. My Testing currently doesn't get this far to see if this is correct or not yet though, unfortunately.

TheLydonKing:
Has anyone been able to test this code or find any glaring issues while I was away? I had a think about this during some time off and I don't understand what else is required to work or how to debug what the issue is, as I stated above there's a lot of different issues that could be present.

api/src/test/scala/za/co/absa/loginsvc/rest/config/provider/ConfigProviderTest.scala
```diff
@@ -48,6 +48,7 @@ class ConfigProviderTest extends AnyFlatSpec with Matchers  {
     activeDirectoryLDAPConfig.order shouldBe 2
     activeDirectoryLDAPConfig.serviceAccount.username shouldBe "CN=svc-ldap,OU=Users,OU=CORP Accounts,DC=corp,DC=dsarena,DC=com"
```
dk1844, line 49: Let just maybe strip the dsarena.com domain from the tests?

dk1844 (approved):
LGTM, nicely done!

api/src/test/scala/za/co/absa/loginsvc/rest/config/provider/ConfigProviderTest.scala
TheLydonKing: Amended

PR #107, opened 2024-11-15
Add SPNEGO authentication header

### LINKED ISSUE #106: Add SPNEGO auth to response headers
Add SPNEGO header so users are more easily aware that it is an option.
Currently, the endpoint only advertises Basic Auth.
spnego-not-correctly-advertized-in-ls 1

TheLydonKing:
Added WWW-Authentication: Negotiate Header to 401 Status response when Kerberos is enabled

closes #106

### DIFF
--- a/api/src/main/scala/za/co/absa/loginsvc/rest/SecurityConfig.scala
+++ b/api/src/main/scala/za/co/absa/loginsvc/rest/SecurityConfig.scala
@@ -26,8 +26,7 @@ import org.springframework.security.web.authentication.www.BasicAuthenticationFi
 import za.co.absa.loginsvc.rest.config.provider.AuthConfigProvider
 import za.co.absa.loginsvc.rest.provider.kerberos.KerberosSPNEGOAuthenticationProvider
 
-import javax.servlet.http.HttpServletRequest
-import javax.servlet.http.HttpServletResponse
+import javax.servlet.http.{HttpServletRequest, HttpServletResponse}
 import org.springframework.security.core.AuthenticationException
 
 @Configuration
--- a/api/src/main/scala/za/co/absa/loginsvc/rest/SecurityConfig.scala
+++ b/api/src/main/scala/za/co/absa/loginsvc/rest/SecurityConfig.scala
@@ -72,6 +72,7 @@ class SecurityConfig @Autowired()(authConfigsProvider: AuthConfigProvider) {
                                          response: HttpServletResponse,
                                          authException: AuthenticationException) => {
                 response.setStatus(HttpServletResponse.SC_UNAUTHORIZED)
+                response.addHeader("WWW-Authenticate", """Basic realm="Realm"""")
                 response.addHeader("WWW-Authenticate", "Negotiate")
               })
         }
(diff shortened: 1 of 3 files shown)


api/src/main/scala/za/co/absa/loginsvc/rest/SecurityConfig.scala
```diff
@@ -58,13 +61,20 @@ class SecurityConfig @Autowired()(authConfigsProvider: AuthConfigProvider) {
...
+          http.addFilterBefore(
               kerberos.spnegoAuthenticationProcessingFilter,
               classOf[BasicAuthenticationFilter])
-          }
+              .exceptionHandling()
+              .authenticationEntryPoint((request: HttpServletRequest,
+                                         response: HttpServletResponse,
+                                         authException: AuthenticationException) => {
+                response.setStatus(HttpServletResponse.SC_UNAUTHORIZED)
+                response.addHeader("WWW-Authenticate", "Negotiate")
```
dk1844, line 75: I tried it and when kerberos was disabled in config, running a POST against the endpoint ( `curl -X POST http://localhost:9090/token/generate -v`) resulted only in:
`< WWW-Authenticate: Basic realm="Realm"`
(as expected ✅ ).

However, when running the same while having kerberos enabled in config, it only returned
`< WWW-Authenticate: Negotiate`
This time I would expect both options being present in the headers:
`< WWW-Authenticate: Basic realm="Realm"`
`< WWW-Authenticate: Negotiate`

dk1844 (commented):
There is also the matter of existing _Swagger v3 OAS Annotations_ (aka OpenAPI docs) - this has bears no information about the Negotiate Auth option.
`GET http://localhost:9090/v3/api-docs.yaml` gives
```
  /token/generate:
    post:
      security:
      - basicAuth: []
  ...
  securitySchemes:
    basicAuth:
      type: http
      scheme: basic
...
```

but it should probably give
```
  /token/generate:
    post:
      security:
      - negotiate: []
      - basicAuth: []
  ...
  securitySchemes:
    negotiate:
      type: http
      scheme: negotiate
    basicAuth:
      type: http
      scheme: basic
```

I think that the solution *might* be adding another
```scala
  @SecurityRequirement(name = "negotiate")
```
and introducing something like
```scala
@SecurityScheme(
  name = "negotiate",
  `type` = SecuritySchemeType.HTTP,
  scheme = "negotiate"
)

I am not sure that it is entirely correct (perhaps do your own research how it should be done properly, but something along these lines

api/src/main/scala/za/co/absa/loginsvc/rest/SecurityConfig.scala
TheLydonKing: Hi Daniel

I see the issue. I was testing and accidentally left -u in the curl command and that would get WWW-Authenticate: Basic realm="Realm" back along with the Negotiate. So it looked like it was working properly on my end. Will sort this out along with the OpenAPI.

dk1844 (approved):
LGTM (tested on my machine)

PR #111, opened 2025-01-08
#45 have a set of keys instead of just one

### LINKED ISSUE #45: Have a set of keys instead of just one
## Background
To allow more UX-friendly keys rotation, having a set of keys for signing JWTs is needed. Once a set of keys is used, the "oldest" one can stop signing keys for time equal to `expiration` of JWT set in config. After that time, it can be replaced with new one.

Example (more smooth rotation):
1. Current keys: [A].
2. New key B is generated (for example by rotation of A). Current keys [B, A].
3. For some time (let's say 30 min) B is not used yet to sign JWT to allow backends based on LS to fetch this new set of keys.
4. After that time, B starts signing JWTs. A becomes "idle" for some time.
5. A gets removed. Current keys: [B].

A set of keys is also useful if we for some reason will have to revoke some JWTs, as we would only have to revoke one key from a set, leaving more users unaffected.

## Feature
Have a set of keys for signing JWTs, each key having its own unique ID, and add this ID to header of JWT.

TheLydonKing:
Release Notes:
- Made the Previously rotated public keys available for clients to fetch and allows for clients to parse JWT's that haven't expired but are using an older keyset.

closes #45

### DIFF
--- a/api/src/main/resources/example.application.yaml
+++ b/api/src/main/resources/example.application.yaml
@@ -7,19 +7,20 @@ loginsvc:
         access-exp-time: 15min
         refresh-exp-time: 9h
         key-rotation-time: 9h
+        key-phase-out-time: 30min
         alg-name: "RS256"
       #Instead of generating the key in memory
       #The Below Config allows for the application to fetch keys from AWS Secrets Manager.
-      aws-secrets-manager:
-        secret-name: "secret"
-        region: "region"
-        private-key-field-name: "privateKey"
-        public-key-field-name: "publicKey"
-        access-exp-time: 15min
-        refresh-exp-time: 9h
-        poll-time: 5min
-        key-phase-out-time: 2min
-        alg-name: "RS256"
+      #aws-secrets-manager:
+        #secret-name: "secret"
+        #region: "region"
+        #private-key-field-name: "privateKey"
+        #public-key-field-name: "publicKey"
+        #access-exp-time: 15min
+        #refresh-exp-time: 9h
+        #poll-time: 5min
+        #key-phase-out-time: 30min
+        #alg-name: "RS256"
     config:
       # Generates git.properties file for use on info endpoint.
       git-info:
--- a/api/src/main/resources/example.application.yaml
+++ b/api/src/main/resources/example.application.yaml
@@ -10,15 +10,16 @@ loginsvc:
         alg-name: "RS256"
       #Instead of generating the key in memory
       #The Below Config allows for the application to fetch keys from AWS Secrets Manager.
-      #aws-secrets-manager:
-        #secret-name: "secret"
-        #region: "region"
-        #private-key-field-name: "privateKey"
-        #public-key-field-name: "publicKey"
-        #access-exp-time: 15min
-        #refresh-exp-time: 9h
-        #poll-time: 5min
-        #alg-name: "RS256"
+      aws-secrets-manager:
+        secret-name: "secret"
+        region: "region"
+        private-key-field-name: "privateKey"
+        public-key-field-name: "publicKey"
+        access-exp-time: 15min
+        refresh-exp-time: 9h
+        poll-time: 5min
+        key-phase-out-time: 2min
+        alg-name: "RS256"
     config:
       # Generates git.properties file for use on info endpoint.
       git-info:
--- a/api/src/main/scala/za/co/absa/loginsvc/rest/config/auth/ServiceAccountConfig.scala
+++ b/api/src/main/scala/za/co/absa/loginsvc/rest/config/auth/ServiceAccountConfig.scala
@@ -90,11 +90,12 @@ case class AwsSecretsLdapUserConfig(private val secretName: String,
     try {
       val secretsOption = AwsSecretsUtils.fetchSecret(secretName, region, Array(usernameFieldName, passwordFieldName))
 
-      if(secretsOption.isEmpty)
+      secretsOption.fold(
         throw new Exception("Error retrieving username and password from from AWS Secrets Manager")
+      ) { secrets =>
+        (secrets(usernameFieldName), secrets(passwordFieldName))
+      }
 
-      val secrets = secretsOption.get
-      (secrets(usernameFieldName), secrets(passwordFieldName))
     } catch {
       case e: Throwable =>
         logger.error(s"Error occurred retrieving account data from AWS Secrets Manager", e)
--- a/api/src/main/scala/za/co/absa/loginsvc/rest/config/auth/ServiceAccountConfig.scala
+++ b/api/src/main/scala/za/co/absa/loginsvc/rest/config/auth/ServiceAccountConfig.scala
@@ -93,7 +93,7 @@ case class AwsSecretsLdapUserConfig(private val secretName: String,
       secretsOption.fold(
         throw new Exception("Error retrieving username and password from from AWS Secrets Manager")
       ) { secrets =>
-        (secrets(usernameFieldName), secrets(passwordFieldName))
+        (secrets.secretValue(usernameFieldName), secrets.secretValue(passwordFieldName))
       }
 
     } catch {
--- a/api/src/main/scala/za/co/absa/loginsvc/rest/config/auth/ServiceAccountConfig.scala
+++ b/api/src/main/scala/za/co/absa/loginsvc/rest/config/auth/ServiceAccountConfig.scala
@@ -88,10 +88,14 @@ case class AwsSecretsLdapUserConfig(private val secretName: String,
 
   private def getUsernameAndPasswordFromSecret: (String, String) = {
     try {
-      val secrets = AwsSecretsUtils.fetchSecret(secretName, region, Array(usernameFieldName, passwordFieldName))
+      val secretsOption = AwsSecretsUtils.fetchSecret(secretName, region, Array(usernameFieldName, passwordFieldName))
+
+      if(secretsOption.isEmpty)
+        throw new Exception("Error retrieving username and password from from AWS Secrets Manager")
+
+      val secrets = secretsOption.get
       (secrets(usernameFieldName), secrets(passwordFieldName))
-    }
-    catch {
+    } catch {
       case e: Throwable =>
         logger.error(s"Error occurred retrieving account data from AWS Secrets Manager", e)
         throw e
--- a/api/src/main/scala/za/co/absa/loginsvc/rest/controller/TokenController.scala
+++ b/api/src/main/scala/za/co/absa/loginsvc/rest/controller/TokenController.scala
@@ -28,7 +28,7 @@ import org.springframework.security.core.Authentication
 import org.springframework.web.bind.annotation._
 import org.springframework.web.server.ResponseStatusException
 import za.co.absa.loginsvc.model.User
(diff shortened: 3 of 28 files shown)


api/src/main/scala/za/co/absa/loginsvc/rest/controller/TokenController.scala
```diff
@@ -145,11 +145,36 @@ class TokenController @Autowired()(jwtService: JWTService) {
...
+      new ApiResponse(responseCode = "200", description = "Payload containing current and previous public keys is returned",
+        content = Array(new Content(
+          schema = new Schema(implementation = classOf[PublicKey]),
+          examples = Array(new ExampleObject(value = "{\n  \"key\": \"ABCDEFGH1234\"\n}")))
+        )
+      )
+    )
+  )
+  @GetMapping(
+    path = Array("/public-key/all"),
```
dk1844, line 166: I would rather suggest `/public-keys` path of this endpoint

api/src/main/scala/za/co/absa/loginsvc/rest/config/auth/ServiceAccountConfig.scala
```diff
@@ -88,10 +88,14 @@ case class AwsSecretsLdapUserConfig(private val secretName: String,
 
   private def getUsernameAndPasswordFromSecret: (String, String) = {
     try {
-      val secrets = AwsSecretsUtils.fetchSecret(secretName, region, Array(usernameFieldName, passwordFieldName))
+      val secretsOption = AwsSecretsUtils.fetchSecret(secretName, region, Array(usernameFieldName, passwordFieldName))
+
+      if(secretsOption.isEmpty)
+        throw new Exception("Error retrieving username and password from from AWS Secrets Manager")
+
+      val secrets = secretsOption.get
```
dk1844, line 96: I think option processing can be done better without calling `Option.get` e.g. via

```scala
      secretsOption.fold(
        throw new Exception("Error retrieving username and password from from AWS Secrets Manager")
      ){ secrets =>
        (secrets(usernameFieldName), secrets(passwordFieldName))  
      }
```

or more obviously via
```scala
      secretsOption match {
        case None => throw new Exception("Error retrieving username and password from from AWS Secrets Manager")
        case Some(secrets) =>
          (secrets(usernameFieldName), secrets(passwordFieldName))
      }
   
```

dk1844 (commented):
- I am wondering if it makes sense to distinguish current and old publicKey for the clients -- whether just not give them a set/list (then we can be generic of what we give them in the future and they just have to trust the public keys as a set -- one of them should be the signing one).
 - I don't see the functionality where the previous publicKey is phased out after some time has passed after the rotation - and I don't see a configuration for these timeouts either

api/src/main/scala/za/co/absa/loginsvc/rest/controller/TokenController.scala
```diff
@@ -145,11 +145,36 @@ class TokenController @Autowired()(jwtService: JWTService) {
...
 
+  @Tags(Array(new Tag(name = "token")))
+  @Operation(
+    summary = "Gives payload with the current and previously rotated RSA256 public key",
+    description = """Alternative to /public-key - exposes current and previous public keys allowing users to verify a JWT after rotation.""",
+    responses = Array(
+      new ApiResponse(responseCode = "200", description = "Payload containing current and previous public keys is returned",
+        content = Array(new Content(
+          schema = new Schema(implementation = classOf[PublicKey]),
+          examples = Array(new ExampleObject(value = "{\n  \"key\": \"ABCDEFGH1234\"\n}")))
```
dk1844, line 160:
```suggestion
          examples = Array(new ExampleObject(value = "{\n  \"key\": \"ABCDEFGH1234\"\n  \"previousKey\": \"ZYXWVUT9876\"\n}"))))
```

I think there is still the old description there

TheLydonKing:
> * I am wondering if it makes sense to distinguish current and old publicKey for the clients -- whether just not give them a set/list (then we can be generic of what we give them in the future and they just have to trust the public keys as a set -- one of them should be the signing one).
>   -> so I would rather keep the `PublicKey` class intact for the current endpoint and created a new class `PublicKeySet` that would serve the new purpose.
> * I don't see the functionality where the previous publicKey is phased out after some time has passed after the rotation - and I don't see a configuration for these timeouts either

- Will do on the PublicKeySet
- Ok, I was under the impression that the previous key would be phased out on rotation but I can add another schedule that phases it out after a set time until rotation occurs.

api/src/main/scala/za/co/absa/loginsvc/rest/service/jwt/JWTService.scala
```diff
@@ -140,13 +143,24 @@ class JWTService @Autowired()(jwtConfigProvider: JwtConfigProvider, authSearchSe
     (refreshedAccessToken, refreshToken)
   }
 
-  def publicKey: PublicKey = keyPair.getPublic
+  def publicKey: (PublicKey, Option[PublicKey]) = {
```
dk1844, line 146: I think this method should now be called `publicKeys`

api/src/main/scala/za/co/absa/loginsvc/rest/service/jwt/JWTService.scala
```diff
@@ -140,13 +143,24 @@ class JWTService @Autowired()(jwtConfigProvider: JwtConfigProvider, authSearchSe
...
+    val currentPublicKey = keyPair._1.getPublic
+    val previousPublicKey = keyPair._2.map(_.getPublic)
+    (currentPublicKey, previousPublicKey)
```
dk1844, line 149: I think it is a good rule of thumb to avoid calling `._1` and `._2` on a tuple -- it is advisable to rather give the fields a name for better readablity.

```suggestion
    val currentPublicKey = keyPair._1.getPublic
    val previousPublicKey = keyPair._2.map(_.getPublic)
    (currentPublicKey, previousPublicKey)
```

Ideally, you may consider applying this rule elsewhere as well.

api/src/main/resources/example.application.yaml
```diff
@@ -10,15 +10,16 @@ loginsvc:
...
-        #alg-name: "RS256"
+      aws-secrets-manager:
+        secret-name: "secret"
+        region: "region"
+        private-key-field-name: "privateKey"
+        public-key-field-name: "publicKey"
+        access-exp-time: 15min
+        refresh-exp-time: 9h
+        poll-time: 5min
+        key-phase-out-time: 2min
```
dk1844, line 21:
```suggestion
        key-phase-out-time: 30min
```
(a more realistic example, but more about the pahse-out later)

api/src/main/scala/za/co/absa/loginsvc/rest/service/jwt/JWTService.scala
```diff
@@ -195,12 +228,25 @@ class JWTService @Autowired()(jwtConfigProvider: JwtConfigProvider, authSearchSe
...
+        keyPair = (keyPair._1, None)
+      }
+    }, phaseOutTime.toMillis, TimeUnit.MILLISECONDS)
+
+    Runtime.getRuntime.addShutdownHook(new Thread(() => {
+      scheduledFuture.cancel(false)
+      this.close()
+    }))
+  }
+
```
dk1844, line 249: In the [comment above](https://github.com/AbsaOSS/login-service/pull/111#issuecomment-2582194805) you wrote
> Ok, I was under the impression that the previous key would be phased out on rotation but I can add another schedule that phases it out after a set time until rotation occurs.

I think you were onto something with this thinking. It probably would have sufficed somewhat (the old key would linger for e.g. a day instead of e.g. 30 min as per `key-phase-out-time`, but it would have worked.

With this scheduler implementation, it works nicely (phase out after `key-phase-out-time`) except for when the service gets restarted -- then the AWSPREVIOUS key appears again and the schedule to phase it out starts anew, unfortunately. Therefore I propose the following:
 - revert as it was with AWSPREVIOUS previous key lingering until next rotation (less preferred but okayish)
 - remove the phase-out scheduling while keeping the `key-phase-out-time` setting. Use this setting to determine if the AWSPREVIOUS is older than now-`key-phase-out-time` while loading it in `AwsSecretsUtils.fetchSecret` - this would be much simpler, I think. (just filtering on loading)

There should be a field `CreatedDate` to be used for this. Evidence from AWS CLI:
```shell
 aws secretsmanager get-secret-value --secret-id afs1-npintdebdtools-dev-bdtools-lsd-jwt-secret --version-stage AWSPREVIOUS--output json
{
    "ARN": "arn:aws:secretsmanager:af-south-1:923318442476:secret:afs1-npintdebdtools-dev-bdtools-lsd-jwt-secret-iXJ5qL",
    "Name": "afs1-npintdebdtools-dev-bdtools-lsd-jwt-secret",
    "VersionId": "d1e7db68-5cd6-4208-b433-ce5e1b733392",
    "SecretString": "{\"privateKey\": \"MIIEvAIBADANB...==\", \"publicKey\": \"MIIBIjANBgk...\"}",
    "VersionStages": [
        "AWSPREVIOUS",
    ],
    "CreatedDate": "2025-01-14T05:11:14.770000+01:00"
}
```

dk1844 (commented):
See the comments, otherwise LGTM 😃

api/src/main/scala/za/co/absa/loginsvc/utils/AwsSecretsUtils.scala
```diff
@@ -61,4 +64,34 @@ object AwsSecretsUtils {
...
+
+    val targetVersionId = versionStage match {
+      case Some(stage) =>
+        versionIdsToStages.find { case (_, stages) => stages.asScala.contains(stage) }.map(_._1)
+      case None =>
+        versionIdsToStages.find { case (_, stages) => stages.asScala.contains("AWSCURRENT") }.map(_._1)
+    }
+
+    targetVersionId.flatMap(_ => Option(describeResponse.createdDate()))
+  }
```
dk1844, line 96: I have a feeling that doing this request separately is unnecessary.

The original
```scala
val getSecretValueResponse: GetSecretValueResponse = client.getSecretValue(getSecretValueRequest)
```
should already hold the created datetime. 

I have tried to debug this call and the result looks like this:
```
GetSecretValueResponse(ARN=arn:aws:secretsmanager:af-south-1:923318442476:secret:bdtools/login-service/login-service-svc-account-secret-5gvnk1, Name=bdtools/login-service/login-service-svc-account-secret, VersionId=acf0767c-7434-4506-b2c6-69b8f91976a8, SecretString=*** Sensitive Data Redacted ***, VersionStages=[AWSCURRENT], CreatedDate=2024-09-18T14:01:11.764Z)
```

Therefore, I believe, one should be able to extract the createdDate from there just with 
```scala
getSecretValueResponse.createdDate()
```
and include this in the resulting value of `fetchSecret` method (I suggest creating a custom case class for it).

api/src/main/scala/za/co/absa/loginsvc/rest/service/jwt/JWTService.scala
jakipatryk: 100% agree

dk1844 (approved):
LGTM

PR #115, opened 2025-01-21
Expiration now based on Current Key Create Time instead of Previous key create time

### LINKED ISSUE #114: Fix isExpired for keyphaseOut
Key Phase expiry is currently being based on the previous keys creation time. This is incorrect as this reflects the previous rotation time and therefore the key will always be expired.

We need to base the expiry time on the creation time of the current key as that's the latest rotation time.

TheLydonKing:
Release Notes:
-  Use Current Key creation time to calculate expiration of keys instead of Previous Key Creation Time.

Minor BugFix for Key Phase expiry being based on the previous keys creation time. This is incorrect as this reflects the previous rotation time and therefore the key will always be expired.

We need to base the expiry time on the creation time of the current key as that's the latest rotation time.

closes #114


dk1844 (approved):
Haha, I did not realize this issue.
Makes sense like this (just read the code)

PR #118, opened 2025-02-06
117 delay option for new set of key signing

### LINKED ISSUE #117: Delay option for new set of key signing
Add new config property and code that delays the signing of new JWT's with the new keys for a set period of time. JWT's within this period will be signed with the previously rotated JWT.

TheLydonKing:
Release Notes:
- Added new Optional Config setting that allows for a layover period before signing new JWT's with the new keys
- Fixed RefreshToken to use both available public keys

closes #117

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,9 @@ AbsaOSS Common Login service using JWT Public key signatures
 To interact with the service, most notable endpoints are
  - `/token/generate` to generate access & refresh tokens
  - `/token/refresh` to obtain a new access token with a still-valid refresh token
- - `/token/public-key` to obtain public key to verify tokens including their validity window
+ - `/token/public-key` to obtain the currently signing public key to verify tokens including their validity window
+ - `/token/public-keys` to obtain all available public keys including the current and previously rotated keys.
+ - `/token/public-key-jwks` gives same data as `/token/public-keys` but in the form of a JSON Web Key Set.
 
 Please, refer to the [API documentation](#api-documentation) below for details of the endpoints.
 
@@ -201,13 +203,27 @@ loginsvc:
          access-exp-time: 15min
          refresh-exp-time: 9h
          key-rotation-time: 9h
+         key-lay-over-time: 15min
+         key-phase-out-time: 30min
          alg-name: "RS256"
 ```
 There are a few important configuration values to be provided:
 - `access-exp-time` which indicates how long an access token is valid for,
 - `refresh-exp-time` which indicates how long a refresh token is valid for,
 - Optional property: `key-rotation-time` which indicates how often Key pairs are rotated. Rotation will be disabled if missing.
+- Optional property: `key-lay-over-time` which indicates a delay after rotation before using the newly created key for signing. Lay-over will be disabled if missing.
+- Optional property: `key-phase-out-time` which indicates the time to phase out the older key. Timer is scheduled after `key-lay-over-time` if enabled. Phase-out will be disabled if missing.
 - `alg-name` which indicates which algorithm is used to encode your keys.
+
+Using the above values, the optional properties will give the following effect after the 1st rotation at 9 hours:
+```
+t=0: keys rotation happens
+t=0-14m: layover period: old key from before rotation is still used for signing. Both public keys available from public-keys endpoint.
+t=15-29m: layover is over: new key from after rotation is used for signing. Both public keys available from public-keys endpoint.
+t=30m+: phase-out happens: new key from after rotation is used for signing. Old Key is no longer available from public-keys endpoint.
+```
+These properties cannot be enabled if rotation is not enabled. The combined values of these properties cannot be higher than the rotation time.
+
 
 To setup for AWS Secrets Manager, your config should look like so:
 ```
@@ -222,6 +238,8 @@ loginsvc:
         access-exp-time: 15min
         refresh-exp-time: 9h
         poll-time: 30min
+        key-lay-over-time: 15min
+        key-phase-out-time: 30min
         alg-name: "RS256"
 ```
 Your AWS Secret must have at least 2 fields which correspond to the above properties:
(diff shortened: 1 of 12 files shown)


api/src/test/scala/za/co/absa/loginsvc/rest/service/jwt/JWTServiceTest.scala
```diff
@@ -294,11 +294,11 @@ class JWTServiceTest extends AnyFlatSpec with BeforeAndAfterEach with Matchers {
     assert(initPublicKey._1 == jwtService.publicKeys._2.orNull)
   }
 
-  it should "phase out older keys after 8 seconds" in {
+  it should "phase out older keys after 30 seconds" in {
```
dk1844, line 297: These times are barely bearable, is it necessary to have such long-taking tests?

dk1844 (commented):
Most of the comments are details, but adding tests to the AwsSecretsManagerKeyConfig seem necessary

api/src/main/resources/example.application.yaml
```diff
@@ -8,6 +8,7 @@ loginsvc:
         refresh-exp-time: 9h
         key-rotation-time: 9h
         key-phase-out-time: 30min
+        key-lay-over-time: 15min
```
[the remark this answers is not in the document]
TheLydonKing: > In other words, do I understand correctly that phaseout is always counted from the rotation, so the actual phaseout's duration is only phaseOut-layover?
> 
> **If I got it right, is this a good idea? Would it be better if the time windows of phaseOut was not dependent on layover?**
> 
> So: you could actually have e.g. phaseOut=15min and layover=30min, and phaseOut would only start after layover finished (so after 45min if layover was defined). This would not force phaseOut > layover, anymore, too.
> 
> However, this is just an idea -- when documented clearly, it is not a problem if it behaves as now, it is just a matter of how the configuration times are defined (just conditional addition/subtracting)

Hi

Yes, actually that makes a lot of sense. So will implement it like this.

api/src/main/resources/example.application.yaml
```diff
@@ -103,7 +104,7 @@ management:
   health:
     ldap:
       # TODO: Enable Ldap check for actuator/health when fully Implemented - issue #34
-      enabled: "false"
+      enabled: "true"
```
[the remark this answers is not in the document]
TheLydonKing: Yeah, this was a mistake, it's changed back

README.md
```diff
@@ -201,14 +203,28 @@ loginsvc:
...
+- Optional property: `key-lay-over-time` which indicates a delay after rotation before using the newly created key for signing. Lay-over will be disabled if missing.
+- Optional property: `key-phase-out-time` which indicates the time to phase out the older key. Timer is scheduled after `key-lay-over-time` if enabled. Phase-out will be disabled if missing.
 - `alg-name` which indicates which algorithm is used to encode your keys.
 
+Using the above values, the optional properties will give the following effect after the 1st rotation at 9 hours:
+```
+t=0: keys rotation happens
+t=0-14m: layover period: old key from before rotation is still used for signing. Both public keys available from public-keys endpoint.
+t=15-29m: layover is over: new key from after rotation is used for signing. Both public keys available from public-keys endpoint.
+t=30m+: phase-out happens: new key from after rotation is used for signing. Old Key is no longer available from public-keys endpoint.
```
dk1844, line 223:
```suggestion
t=15-44m: layover is over: new key from after rotation is used for signing. Both public keys available from public-keys endpoint.
t=45m+: phase-out happens: new key from after rotation is used for signing. Old Key is no longer available from public-keys endpoint.
```

Since now you have made phase-out window to be counted *after* the layover is concluded.

README.md
```diff
@@ -236,7 +254,17 @@ There are a few important configuration values to be provided:
...
 - Optional property:`poll-time` which indicates how often key pairs (`private-key-field-name` and `public-key-field-name`) are polled and fetched from AWS Secrets Manager. Polling will be disabled if missing.
+- Optional property: `key-lay-over-time` which indicates a delay after rotation before using the newly created key for signing. Lay-over will be disabled if missing.
+- Optional property: `key-phase-out-time` which indicates the time to phase out the older key. Timer is scheduled after `key-lay-over-time` if enabled. Phase-out will be disabled if missing.
 - `alg-name` which indicates which algorithm is used to encode your keys.
+  Using the above values, the optional properties will give the following effect after the 1st rotation at 9 hours:
+```
+t=0: keys rotation happens
+t=0-14m: layover period: old key from before rotation is still used for signing. Both public keys available from public-keys endpoint.
+t=15-29m: layover is over: new key from after rotation is used for signing. Both public keys available from public-keys endpoint.
+t=30m+: phase-out happens: new key from after rotation is used for signing. Old Key is no longer available from public-keys endpoint.
```
dk1844, line 265:
```suggestion
t=15-44m: layover is over: new key from after rotation is used for signing. Both public keys available from public-keys endpoint.
t=45m+: phase-out happens: new key from after rotation is used for signing. Old Key is no longer available from public-keys endpoint.
```

Same as above, I think

api/src/test/scala/za/co/absa/loginsvc/rest/config/jwt/AwsSecretsManagerKeyConfigTest.scala
```diff
@@ -60,17 +89,104 @@ class AwsSecretsManagerKeyConfigTest extends AnyFlatSpec with Matchers {
...
+    when (mockSecretsUtil.fetchSecret(anyString(),anyString(),any[Array[String]](),eqMatch(None)))
+      .thenReturn(currentSecret)
+    when(mockSecretsUtil.fetchSecret(anyString(),anyString(),any[Array[String]](),eqMatch(Some("AWSPREVIOUS"))))
+      .thenReturn(None)
+
+    val keysDuringLayover = awsSecretsManagerKeyConfig.fetchKeySetsFromCloud(mockSecretsUtil)
+
+    assert(keysDuringLayover._1.getPrivate == currentKeyPair.getPrivate)
+    assert(keysDuringLayover._1.getPublic == currentKeyPair.getPublic)
+    assert(keysDuringLayover._2.isEmpty)
```
dk1844, line 128:
```suggestion
    val (currentKey, optPreviousKey) = awsSecretsManagerKeyConfig.fetchKeySetsFromCloud(mockSecretsUtil)

    assert(currentKey.getPrivate == currentKeyPair.getPrivate)
    assert(currentKey.getPublic == currentKeyPair.getPublic)
    assert(optPreviousKey.isEmpty)
```

This has been already mentioned but reiterating - please try not to use `._1` `._2` unless warrantied -- this is harder to read. Instead, please use named values to better signify their meaning. This makes the code much easily digestable.

(feel free to apply else where)

api/src/main/scala/za/co/absa/loginsvc/utils/AwsSecretsUtils.scala
```diff
@@ -66,3 +66,12 @@ object AwsSecretsUtils {
     }
   }
 }
+
+trait SecretUtil {
```
dk1844, line 70:
```suggestion
trait SecretUtils {
```

suggesting a rename to follow the suite of the object name

api/src/main/scala/za/co/absa/loginsvc/rest/config/jwt/AwsSecretsManagerKeyConfig.scala
```diff
@@ -36,18 +36,49 @@ case class AwsSecretsManagerKeyConfig(
...
+        .map(_ => ConfigValidationSuccess)
+        .getOrElse(ConfigValidationError(ConfigValidationException("publicKeyFieldName is empty"))),
+    )
+
+    val awsSecretsResultsMerge = awsSecretsResults.foldLeft[ConfigValidationResult](ConfigValidationSuccess)(ConfigValidationResult.merge)
+
+    super.validate().merge(awsSecretsResultsMerge)
+  }
+
+  def fetchKeySetsFromCloud(secretsUtils: SecretUtil = AwsSecretsUtils): (KeyPair, Option[KeyPair]) = {
```
dk1844, line 75:
```suggestion
/**
   * Some description here
   * @param secretsUtils
   * @return tuple of [AWSCURRENT keyPair, optionally AWSPREVIOUS]
   */
  private[jwt] def fetchKeySetsFromCloud(secretsUtils: SecretUtil = AwsSecretsUtils): (KeyPair, Option[KeyPair]) = {
```

Suggesting: limiting visibility (essentially private but available for tests), 
simple doc to signify which key is which (since they are of the same type)

dk1844 (commented):
Just some details (code quality/readability improvements)

dk1844 (approved):

PR #130, opened 2025-07-07
Experimental/129 get generate negotiate

### LINKED ISSUE #129: PoC: attempt to do chrome negotiation on a get enpoint
copy POST /token/generate -> GET /experimental/token/generate

Attemp to reach /experimental/token/generate with Chrome with active kerberos ticket (negotiate)

@TheLydonKing

dk1844:
This PR shows an experimental GET endpoint at `/token/experimental/get-generate` that supports 
1 spnego negotiate from browser
2 basic auth

This endpoint will only work if config allows it - via `loginsvc.rest.experimental.enabled=true` (otherwise throws 409)

## Test run:
Tested to work at
https://login-service-dev.npintdebdtools.aws.dsarena.com/token/experimental/get-generate

with Edge and Chromium.
### Chrome with `AuthNegotiateDelegateAllowlist` set
**Does not currently work in Chrome** because it has `AuthNegotiateDelegateAllowlist` -- and LS is not in the allowed list. (We may try to add LS endpoint to be included)

Release notes:
 - experimental GET endpoint at `/token/experimental/get-generate` added (must be allowed in config via `loginsvc.rest.experimental.enabled=true`)

Closes #129


TheLydonKing (approved):
LGTM

PR #136, opened 2025-09-01
Remove jacoco forks

miroslavpojer:
Release Notes:
- Used new GH Action for Jacoco reports.
- Used Jacoco agent solution instead of sbt-jacoco plugin for measuring code coverage.
- Used new library for method filtration for jacoco reports.


build.sbt
```diff
@@ -14,7 +14,7 @@
  */
 
 import Dependencies._
-import com.github.sbt.jacoco.report.JacocoReportSettings
+//import com.github.sbt.jacoco.report.JacocoReportSettings
```
dk1844, line 17: The comments here are examples or to be cleaned up?

PR #153, opened 2025-12-01
(#151) Token Generation Case Sensitivity option added to Client Library

### LINKED ISSUE #151: token/generate case sensitivity for clientLibrary
## Background
New feature of case-sensitivity-driven groups filtering has been added to the backend in #146
This feature should also be present in the client library API

## Feature
Add an option to explicitly add case-sensitivity flag to the token/generate call in the client library API

## Business Value
Clients using the client library should be able to get the pre-v2 behavior that is actually available on the backend

TheLydonKing:
Release Notes:
- Added Case Sensitivity option for client library

closes #151

### DIFF
--- a/examples/src/main/scala/za/co/absa/clientexample/Application.scala
+++ b/examples/src/main/scala/za/co/absa/clientexample/Application.scala
@@ -91,9 +91,9 @@ object Application {
       try {
         val (accessToken, refreshToken) = authMethod match {
           case "1" =>
-            tokenRetriever.fetchAccessAndRefreshToken(username, password, List.empty)
+            tokenRetriever.fetchAccessAndRefreshToken(username, password, List.empty, false)
           case "2" =>
-            tokenRetriever.fetchAccessAndRefreshToken(None, None, List.empty)
+            tokenRetriever.fetchAccessAndRefreshToken(None, None, List.empty, false)
         }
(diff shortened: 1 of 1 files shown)


clientLibrary/src/main/scala/za/co/absa/loginclient/tokenRetrieval/client/TokenRetrievalClient.scala
```diff
@@ -58,10 +63,15 @@ case class TokenRetrievalClient(host: String) {
...
+   * @param caseSensitiveGroups A boolean indicating whether the group prefixes should be treated as case sensitive.
    * @return An AccessToken object representing the retrieved access token (JWT) from the login service.
    */
-  def fetchAccessToken(keytabLocation: Option[String], userPrincipal: Option[String], groups: List[String]): AccessToken = {
-    fetchAccessAndRefreshToken(keytabLocation, userPrincipal, groups)._1
+  def fetchAccessToken(
+    keytabLocation: Option[String],
+    userPrincipal: Option[String],
+    groups: List[String],
+    caseSensitiveGroups: Boolean): AccessToken = {
```
dk1844:
```suggestion
    caseSensitiveGroups: Boolean = false): AccessToken = {
```

PR #155, opened 2025-12-03
(#154) Set Groups and Case Sensitivity to optional Parameters

### LINKED ISSUE #154: Set Groups and Case sensitivity for token retrieval in client lib as optional
We currently let the user set the Groups and Case sensitivity triggers for the URI in the client lib but currently they are mandatory to set. We need to make them optional as they are both optional on the API itself.

Current issue: Setting the as optional in the current state creates ambiguity for the compiler as we are using overloaded methods (One for basic auth and one for kerberos)

Solution: Create an authentication sealed trait and utilize that in the methods instead of using 2 overloaded methods.

TheLydonKing:
Release Notes:
- Removed overloaded methods and introduced AuthMethod trait to indicate authentication method. This removes ambiguity with the default parameters and avoids compilation errors.

Closes: #154


clientLibrary/src/main/scala/za/co/absa/loginclient/tokenRetrieval/client/TokenRetrievalClient.scala
```diff
@@ -39,116 +39,50 @@ case class TokenRetrievalClient(host: String) {
...
-   * This method requests an access token (JWT) from the login service using SPNEGO.
-   * This Token is used to access resources which utilize the login Service for authentication.
-   *
-   * @param keytabLocation  Optional location of the keytab file.
-   * @param userPrincipal   Optional userPrincipal name included in the above keytab file.
-   * @param groups          An optional list of PAM groups. If provided, only JWTs associated with these groups are returned if the user belongs to them.
-   * @param caseSensitiveGroups A boolean indicating whether the group prefixes should be treated as case sensitive.
-   * @return An AccessToken object representing the retrieved access token (JWT) from the login service.
-   */
   def fetchAccessToken(
```
dk1844, line 70: This being a library, we should not just remove API as if it was never there between versions.
The existing API should stay in place with `@deprecated` annotation (ideally internally calling the new methods and thus be ready for removal in the future).

clientLibrary/src/main/scala/za/co/absa/loginclient/tokenRetrieval/client/TokenRetrievalClient.scala
```diff
@@ -39,116 +39,50 @@ case class TokenRetrievalClient(host: String) {
...
-   * This method requests a refresh token from the login service using the specified username and password.
-   * This token may be used to acquire a new access token (JWT) when the current access token expires.
-   *
-   * @param username The username used for authentication.
-   * @param password The password associated with the provided username.
-   * @return A RefreshToken object representing the retrieved refresh token from the login service.
-   */
-  def fetchRefreshToken(username: String, password: String): RefreshToken = {
-    fetchAccessAndRefreshToken(username, password, List.empty, false)._2
+  def fetchRefreshToken(authMethod: AuthMethod): RefreshToken = {
```
dk1844, line 85: Neither the new method nor the existing (which should be present for some time as described above) seem to be covered with unit tests.

That is unfortunate, because the appropriate class `TokenRetrievalClientTest` already exists and it would be nice to have it covered. I see two options:

1. You may need to mock the creation of `RestTemplate` & `KerberosRestTemplate` (could be done by moving the creation to a method that you override in the test) to write the unit test properly. 
2. A half-measure would be just to mock the `fetchToken` methods in the class to check if those are receiving expected parameters. Not a complete unit test, but even this would be better than nothing if the option 1 cannot be reasonably done (I believe it should be).

dk1844 (commented):
The code is nice, but with a library we should meet additional standards (see comments below)

clientLibrary/src/main/scala/za/co/absa/loginclient/tokenRetrieval/client/TokenRetrievalClient.scala
TheLydonKing: Ah, will do

dk1844: Nicely done, thanks

clientLibrary/src/test/scala/za/co/absa/loginclient/tokenRetrieval/client/TokenRetrievalClientTest.scala
```diff
@@ -1,51 +1,189 @@
...
+        HttpMethod.POST,
+        requestEntity,
+        classOf[String]
+      )
+      val jsonObject = JsonParser.parseString(response.getBody).getAsJsonObject
+      (
+        AccessToken(jsonObject.get("token").getAsString),
+        RefreshToken(jsonObject.get("refresh").getAsString)
+      )
+    }
```
dk1844: Since you are already providing the mocked exchange templates, I assume you want the full solution. Good choice.

However, you do not need to override all the `fetchToken` and `refreshAccessToken` (also these should be rather tested, not overridden). I think all you need is this:

```suggestion
  class MockableTokenRetrievalClient(
    host: String,
    restTemplate: RestTemplate,
    kerberosRestTemplate: KerberosRestTemplate) extends TokenRetrievalClient(host) {

    override def createRestTemplate(): RestTemplate = restTemplate
    override def createKerberosRestTemplate( /* opt params or two methods? */ ): KerberosRestTemplate = kerberosRestTempl
```

while updating the with

```scala
case class TokenRetrievalClient(host: String) {

  private[client] def createKerberosRestTemplate( /* opt params or two methods? */ ): KerberosRestTemplate = {
    new KerberosRestTemplate( /* opt params?*/)
  }

  private[client] def createRestTemplate(): RestTemplate = {
    new RestTemplate()
  }

// and wherever the `new RestTemplate()` or `new KerberosRestTemplate(...)` is used within `TokenRetrievalClient`, you replace it with the methods above.
```

This way, you only override the very minimal to get your exchange mocks in -- while the rest of the `TokenRetrievalClient` stays original as is correctly subjected to unit testing.

clientLibrary/src/main/scala/za/co/absa/loginclient/tokenRetrieval/client/TokenRetrievalClient.scala
```diff
@@ -38,6 +39,25 @@ case class TokenRetrievalClient(host: String) {
 
   private val logger: Logger = LoggerFactory.getLogger(this.getClass)
 
+  private[client] def createKerberosRestTemplate(
+    keyTabLocation: Option[String] = None,
+    userPrincipal: Option[String] = None): KerberosRestTemplate = {
+    (keyTabLocation, userPrincipal) match {
+      case (Some(_), Some(_)) =>
+        logger.info(s"Fetching token with user $userPrincipal")
+        new KerberosRestTemplate(keyTabLocation.get, userPrincipal.get)
```
dk1844:
```suggestion
      case (Some(definedKeyTabLocation), Some(definedUserPrincipal)) =>
        logger.info(s"Fetching token with user $definedUserPrincipal")
        new KerberosRestTemplate(definedKeyTabLocation, definedUserPrincipal)
```

You do not have to resort to `optionValueName.get` this way 😄

clientLibrary/src/test/scala/za/co/absa/loginclient/tokenRetrieval/client/TokenRetrievalClientTest.scala
```diff
@@ -1,51 +1,189 @@
...
+    val mockRestTemplate = mock(classOf[RestTemplate])
+    val mockKerberosTemplate = mock(classOf[KerberosRestTemplate])
+    val tokenResponse = """{"token": "mock-access-token", "refresh": "mock-refresh-token"}"""
+
+    when(mockRestTemplate.exchange(
+      anyString(),
+      any[HttpMethod],
+      any[HttpEntity[String]],
+      any[Class[String]]
+    )).thenReturn(new ResponseEntity(tokenResponse, HttpStatus.OK))
```
dk1844, line 41: Unfortunately, this is not good. This is the place where we express our expectations of what will be sent to the endpoint -- and the expectations must hold for the successful test (also for the test to make sense). If we allow anything, what is the point of the test -- it would just check that there was _something_ sent (that is not enough).

```scala
import org.mockito.ArgumentMatchers.{eq => eqTo} // not to clash with `Matchers.eq`
// this may be common for the whole test or even with hardcoded values
val base64Credentials = new String(Base64.getEncoder.encodeToString((s"$dummyUser:$dummyPassword").getBytes()))

```

```suggestion
  val expectedBasicAuthHeader = new HttpHeaders()
  expectedBasicAuthHeader.set("Authorization", s"Basic $base64Credentials")

    when(mockRestTemplate.exchange(
      eqTo(s"$dummyUri/token/generate"),
      eqTo(HttpMethod.POST),
      eqTo(new HttpEntity[String](null, expectedBasicAuthHeader)),
      eqTo(classOf[String])
    )).thenReturn(new ResponseEntity(tokenResponse, HttpStatus.OK))
```

... and I believe this is needed for every `when(mockRestTemplate.exchange(...))` in this test -- of course the content will vary if it is a `generate` request, `refresh` request`, basic auth vs kerberos etc.

dk1844 (commented):
Good advancement, just one more important thing in the tests:

```scala
when(mockRestTemplate.exchange(
      anyString(),
      any[HttpMethod],
      any[HttpEntity[String]],
      any[Class[String]]
    ))
```
is just too broad for a meaningful check. Please consider doing a better arguments check.

clientLibrary/src/test/scala/za/co/absa/loginclient/tokenRetrieval/client/TokenRetrievalClientTest.scala
TheLydonKing: So, my assumption here is that the Template in this case is mimicking the actual API response and the API is a separate entity from testing the Client entity itself, so irrelevant to testing the client.

I will change it accordingly though.

dk1844: Well, the API is a different entity, but you are testing the client to make sure that the behavior of the API and the client match.

From another angle, class under test is `TokenRetrievalClient` and its public methods are put to the test. Since the methods actually construct the API requests with path, auth and other properties -- checking of these should appear in the unit test. Which, now, it is being done 🤝

dk1844 (approved):
LGTM (read the code, ran tests)

PR #157, opened 2026-01-12
#156 fixing: generic 401 bears no error body content

### LINKED ISSUE #156: Bug: POST token/generate returns empty on 401 while Swagger defines JSON
The swagger UI wrongly indicates a json to be returned for **POST token/generate** in case of **401**, where in reality no response is returned.
See Swagger page

As a result the client code generated from Swagger definition fails upon receiving 401. 

## To Reproduce
Use Swagger UI to execute POST /token/generate with wrong credentials
OR use curl with wrong credentials

## Expected behavior
Return json with error message. Or update Swagger definition not to state application/json.

dk1844:
Fixes #156.

In the particular case, the Error code was sent, but there was no body, now there is a json error body as per Swagger doc.

Release notes:
 - Bugfix: 401 Responses were missing body with error description despite the API doc stating so.


TheLydonKing (approved):
LGTM

dk1844:
FYI @DzMakatun @wajda

PR #158, opened 2026-03-11
feat: MS Entra Bearer token authentication (#127)

### LINKED ISSUE #127: MS Entra integraton
## Background
There are application in our ecosystem that currently use MS Entra authentication.
We would like to give an option to use this ecosystem as well to LS users.

# Tasks
- [ ] #128

oto-macenauer-absa:
## Summary

Adds MS Entra (Azure AD) as a new login mechanism. Users holding a valid Entra Bearer JWT can now exchange it for a login-service access+refresh token pair alongside existing Basic Auth, LDAP and Kerberos methods.

## Changes

- **`MsEntraConfig`** — PureConfig case class (`tenant-id`, `client-id`, `audience`, `order`, optional `attributes` map); implements `ConfigValidatable`
- **`MsEntraTokenValidator`** — validates Entra JWTs via OIDC discovery + Nimbus JOSE; JWKS cached 1 h via Guava `LoadingCache`; injectable `JWKSource` override for tests (no HTTP calls)
- **`MsEntraBearerTokenFilter`** — Spring `OncePerRequestFilter`; intercepts `Authorization: Bearer` headers; populates `SecurityContext` on success, returns HTTP 401 on failure; skips when context already populated
- **`SecurityConfig`** — conditionally registers the filter before `BasicAuthenticationFilter` when `entra.order > 0`
- **`AuthConfigProvider` / `ConfigProvider`** — `getMsEntraConfig` plumbing
- **`Dependencies`** — Guava `CacheBuilder` added to `apiDependencies`
- **`example.application.yaml`** — commented `entra:` config block added

## Tests

Three new test files, all 153 tests pass:
- `MsEntraConfigTest` — validation logic
- `MsEntraTokenValidatorTest` — real RSA key pair, injected `ImmutableJWKSet`, no HTTP
- `MsEntraBearerTokenFilterTest` — Mockito mock validator, `MockHttpServletRequest/Response`

## Release notes:
- Add MS Entra login mechanism

## Configuration

To enable, add to `application.yaml`:
```yaml
loginsvc.rest.auth.provider:
  entra:
    order: 3
    tenant-id: "<tenant-id>"
    client-id: "<client-id>"
    audience: "api://<client-id>"
```

Closes #127


oto-macenauer-absa:
Updated this PR to normalize Graph-resolved Entra usernames to lowercase `samAccountName` values without the domain prefix.

What changed:
- Graph-backed Entra resolution now returns `ab006hm` instead of `CORP\AB006HM`
- configured `domains` entries remain the allow-list for recognized `onPremisesDomainName` values
- successful resolution still logs the mapped AB/NetBIOS value for traceability
- tests/docs were updated to match the new behavior

What this allows:
- users from allowed on-prem domains can authenticate with the username format expected by the local environment
- known Graph domains are still validated before accepting the resolved username
- troubleshooting keeps the domain-to-AB mapping visible in logs without putting the prefix into the token username

dk1844 (commented):
LGTM so far

I would like to test it with our LS dev with our MS Entra AppId + Object ID

api/src/main/scala/za/co/absa/loginsvc/rest/provider/entra/MsEntraGraphClient.scala
```diff
@@ -54,15 +54,21 @@ trait GraphUsernameResolver {
...
+) extends GraphUsernameResolver {
 
   private val logger = LoggerFactory.getLogger(classOf[MsEntraGraphClient])
 
   private val tokenEndpoint =
-    s"https://login.microsoftonline.com/${config.tenantId}/oauth2/v2.0/token"
+    tokenEndpointOverride.getOrElse(s"https://login.microsoftonline.com/${config.tenantId}/oauth2/v2.0/token")
 
-  private val graphUsersBaseUrl = "https://graph.microsoft.com/v1.0/users"
+  private val graphUsersBaseUrl = graphUsersBaseUrlOverride.getOrElse("https://graph.microsoft.com/v1.0/users")
```
dk1844: I strongly believe that these should not be in code. I know these are fallbacks but even so.

oto-macenauer-absa: Addressed in 3d1de53. All hardcoded Microsoft endpoint URLs have been moved into `MsEntraConfig` as two optional fields with public-cloud defaults:

- `loginBaseUrl` (default: `https://login.microsoftonline.com`)
- `graphBaseUrl` (default: `https://graph.microsoft.com`)

All derived URLs (token endpoint, Graph users path, Graph scope, OIDC discovery URL, expected JWT issuer) are now computed from these config values. The test constructor overrides (`tokenEndpointOverride`/`graphUsersBaseUrlOverride`) have been removed — tests now pass custom base URLs via `MsEntraConfig` directly. The `example.application.yaml` documents both fields.

dk1844 (approved):
LGTM (read the code). Looking forward to start experimenting with it.

PR #160, opened 2026-04-21
Add browser-based login form example using Entra ID and MSAL.js

oto-macenauer-absa:
## Summary

Adds a self-contained browser UI example at `examples/ui/login-form/` demonstrating the complete Entra ID → Login Service token-exchange flow.

## Changes

- **`src/auth.js`** — MSAL authentication module (login, silent token acquisition, logout)
- **`src/login-service-client.js`** — HTTP client for the Login Service `/token/generate` endpoint
- **`src/ui.js`** — all DOM manipulation helpers (status, loading state, token display, copy-to-clipboard)
- **`src/app.js`** — entry point wiring the modules together and registering event listeners
- **`index.html`** — single-page UI; loads `config.js` then `dist/bundle.js` (defer)
- **`config.js`** — runtime configuration template (tenantId, clientId, scopes, loginServiceHost)
- **`server.js`** — minimal Express static server (required for MSAL redirect URIs)
- **`package.json`** — build + start scripts; esbuild bundles `src/app.js` → `dist/bundle.js` (IIFE); `prepare` hook auto-builds on `npm install`
- **`README.md`** — project structure, module responsibilities, prerequisites, config and running instructions
- **`.gitignore`** — excludes `node_modules/` and `dist/`

## How to run

```bash
cd examples/ui/login-form
npm install   # installs deps and builds dist/bundle.js
npm start     # serves on http://localhost:3000
```

## Release notes:
* Include an example login form to demonstrate Login Service integration with entra

Closes #159

PR #162, opened 2026-04-23
Bugfix/161 entra timeouts

### LINKED ISSUE #161: MS Entra plugin times out sometimes
## Describe the bug

```
2026-04-23 08:44:10.255  WARN 1 --- [nio-8080-exec-4] z.c.a.l.r.p.entra.MsEntraTokenValidator  : Entra token validation failed: Couldn't retrieve remote JWK set: Read timed out
2026-04-23 08:44:10.255  WARN 1 --- [nio-8080-exec-4] z.c.a.l.r.p.e.MsEntraBearerTokenFilter   : Entra Bearer token rejected: Couldn't retrieve remote JWK set: Read timed out
```

## To Reproduce
Setup working ms entra
1. Request token generation,
4. See error show above (sometimes)

## Expected behavior
Provide means to avoid the timeout in real-life scenarios.
Possibly make the timeout configurable

## Business Value
Timeouts prevent users' login with MS Entra token

dk1844:
Fixes #160 with overridable config values (provided sensible defaults of 10s).

Tested on DEV, got a LS token from Entra token.

Release notes:
 - Entra timeouts are now 10s/10s by default, configurable by `loginsvc.rest.auth.provider.entra.{jwks-connect-timeout-ms, jwks-read-timeout-ms}`

Closes #161

### DIFF
--- a/README.md
+++ b/README.md
@@ -230,14 +230,14 @@ Example config:
 ```
 
 Available options:
-- `order` — provider ordering; `0` disables, `1+` enables.
+- `order` — provider ordering; `0` disables, `1`+ enables.
 - `tenant-id` — Azure AD tenant (directory) ID.
 - `client-id` — App registration (client) ID.
 - `client-secret` — Client secret used to call MS Graph API.
 - `audiences` — List of accepted JWT `aud` values. Empty list accepts any token from the tenant.
 - `domains` — Map of on-premises DNS domains to short names (e.g. `corp.example.com: "CORP"`).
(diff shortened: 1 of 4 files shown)


ABLL526 (approved):
Seems good to me, but 10 seconds as default seems quite long. Any reason for that?

dk1844:
> Seems good to me, but 10 seconds as default seems quite long. Any reason for that?

Well, the original default of 500ms was tested to be too short in our DEV env, so this is just to make sure that we have enough. I hardly see any downside - the users rather want to be logged in while waiting slightly longer than not at all. It is still a overridable default, though.

PR #168, opened 2026-05-14
Feature/167 groups from ldap for entra

### LINKED ISSUE #167: Allow groups slipstream in token/generate (e.g. for MSEntra)
## Background
MSEntra-provided token may lack groups (e.g. they are not configured or present) - the groups may be too big to fit the header etc.

## Feature
Provide a way to fill-in groups from ldap if enabled for this case

## Business Value
There are users who rely on groups being present.
With Entra-provided tokens, they may lack those.

dk1844:
This PR introduces a new feature described in #167.

With `loginsvc.rest.jwt.{aws-secrets-manager|generate-in-memory}.allow-providers-to-refresh-groups-on-generate = false`, the token without groups (typically coming from MS Entra) may be enriched by groups coming from a `UserSearchService` (typically LDAP, but ConfigUsers would work, too).

Necessary changes added:
 - groups filtering has moved to JWTService from TokenController (because we need to filter the fetched groups, too)
 - supported by new unit tests
 - UI of the login form extended to optionally support groups-prefixes (vibe-coded 😆)
 

Release notes:
 - config now must contain ``loginsvc.rest.jwt.{aws-secrets-manager|generate-in-memory}.allow-providers-to-refresh-groups-on-generate = false|true`. If enabled, this bring the ability to fill empty groups for users (obtained by MS Entra).

### Developer testing
Login form UI result with custom groups-prefixes used via MS Entra:
(backend setup includes `allow-providers-to-refresh-groups-on-generate = false`, but empty groups were fetched in the MS Entra JWT)
image
Resulting LS access token contains groups now fetched from LDAP based on the `sub` in the MS Entra JWT:
image

Closes #167


api/src/main/scala/za/co/absa/loginsvc/rest/controller/TokenController.scala
```diff
@@ -135,13 +134,12 @@ class TokenController @Autowired()(jwtService: JWTService, experimentalConfigPro
     }
     val groupPrefixesStrScala = groupPrefixes.toScalaOption
 
-    val filteredGroupsUser = user.applyIfDefined(groupPrefixesStrScala) { (user: User, prefixesStr: String) =>
-      val prefixes = prefixesStr.trim.split(',')
-      user.filterGroupsByPrefixes(prefixes.toSet, caseSensitive)
-    }
+    // groups filtering is pushed down to JWTService as it may be needed for entra-users without groups to have groups added and filtered in the same way
```
dk1844, line 136: This is the get-experimental endpoint.
The fate of it is not decided yet, therefore no generalization effort with the normal post-based `token/generate` (identical impl now).

This feature is disabled on PROD.

TheLydonKing (approved):
LGTM.

